PR #163, opened 2014-08-08
VTK6

gpasquie:
I was kind of tired of refactoring and documenting so I had a look to see if it was feasible, and I think we are almost there!
First I removed some unused files in medVtkInria, and some files that are now available with itkVtkGlue (don't forget to rebuild ITK with this option). Then I basically replaced SetInput with SetInputData/Connection, and it was almost it.
There are still some small issues (the view is not correctly centered sometimes  and there is a crash when opening spherical harmonics) but as far as I tested, it seems ok otherwise.

Guillaume


ocommowi:
#250 is more up to date, closing

PR #187, opened 2014-10-31
Memory improvements (The return)

Hakim-F:
Before  : 
- When we add a layer, the image added is casted into the type of the layer 0 for VR visualization. Memory is therefore allocated, even when VR is not activated. This can be troublesome with 600x600x200 images for example (CTscan) and paint segmentation. You can easily reach 1 Go with no effort.
- The vtkImageAppendComponents  is used to append the different layers into a single vtkImageData which is then setted as an input in the VolumeMapper. The VR visualization of this data can be extremely painful when we have more than 1 layer. We noticed that it is quite slow even when the visibility of all layers except the layer 0 is set to off.

Now: 
- We keep in mind which layers need to be casted, and we cast them only when the VR mode is activated.
- We have changed the code so that when the visilibility of a layer is set to off, its data is not given to the vtkImageAppendComponents to be appended. This enables us to improve interactions, and to be able to have normal VR on a single layer if all the other are Off (visibility) .

Remaining issues : 
- After deleting a layer, no memory seems to be deallocated.
- In a simple case, like for example using paint on an image, a new layer is added, if the user wants to go to VR mode to visualize its layer 0, this will lead to memory allocated for its layer 1 which he probably doesnt want to see in VR, and his/her computer getting slow because of multiple layers being ON (visibility). A possible solution would be to set to Off (visibility) all layers excepts the layer 0 when going to VR. The user can still activate them back, if he wants to visualize them. What do you think ??

### DIFF
--- a/src/medVtkInria/vtkImageView/vtkImageView3D.cxx
+++ b/src/medVtkInria/vtkImageView/vtkImageView3D.cxx
@@ -300,18 +300,21 @@ void vtkImageView3D::SetVolumeMapperToDefault()
 //----------------------------------------------------------------------------
 void vtkImageView3D::SetVolumeRayCastFunctionToComposite()
 {
+  this->castLayers();
   this->VolumeMapper->SetBlendModeToComposite();
 }
 
 //----------------------------------------------------------------------------
 void vtkImageView3D::SetVolumeRayCastFunctionToMaximumIntensityProjection()
 {
+  this->castLayers();
   this->VolumeMapper->SetBlendModeToMaximumIntensity();
 }
 
 //----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


mathildemerle:
Some notes, i tested this old PR, adapted to the current medInria3.4.x branch which have been heavily changed in the way data are displayed in a view.

My test branch https://github.com/medInria/medInria-public/compare/medInria3.4.x...mathildemerle:HakimRAM?expand=1

I didn't see noticeable memory gain between 3.4 and these changes: opening a data, closing a data, opening again, switching to 3D orientation (MSR), switching to VR, closing the data and closing the app.

Perhaps there is a change in the way data are shown in VR, but for now i did not find a difference with medInria3.4.x for the user.

PR #194, opened 2014-11-26
Restore thumbnail in browser area.

rdebroiz:


ocommowi:
Works well but it creates problems with the sizing and spacing of other toolboxes in the left column of the browser...

PR #203, opened 2014-12-11
Remove RGB(A) image support in writer that don't actually support it

fvichot:
This fixes 2 problems : the viewer can actually display uchar vector 3 as RGB but there was a typo in the identifier, so they weren't properly recognised. Secondly, the Meta Image, Gipl and Nifti writers were claiming to handle RGB/RGBA when they actually degrade them to uchar vector 3, so I disabled support for writing RGB/RGBA files with these. Which means now RGB/RGA images are saved as Nrrd, which is the only writer we have that actually saves them properly as RGB/RGBA.

This triggers another, more general question of how to handle vector images... a vector uchar 3 is now currently being interpreted as a RGB image, but it could just as well be a vector field image with uchar vector coordinates. What do we think is the best way to deal with ambiguities such as these ? There should probably be some way on load/import of ambiguous type to be able to tag them with some extra semantic sense.


mikejbuckingham:
Looks good for me!

PR #205, opened 2014-12-12
Filter menu navigator

ocommowi:
This pull request solves a long running bug in the filter toolbox of the browser area. In addition, it tries to add a filter menu in the workspaces navigator. The problem being, that it is probably too large for that small column...


rdebroiz:
Good for me. Just one thing, i think that only one search field  by default would be better.
(need to be rebase on the master to be teseted.)

mathildemerle:
This PR can be closed. This new one is updated: https://github.com/medInria/medInria-public/pull/371 (so, it appears that we applied on music this PR a long time ago)

PR #207, opened 2014-12-15
Refactor of DbImporters/Changes in DB access

mikejbuckingham:
Removed derived importers : Database and NonPersistent. We now only have one importer class, preliminarily named medImporter.

Changed how we import data : controllers are no longer responsible for the creation of the importer, and are instead passed as an argument. Importer doesn't take ownership as the controllers are singletons.

All access to the DB should (in theory) go via the DataManager now. There are a few remaining cases where this isn't true, but I want to analyse how to handle them first.

Please give feedback - I'm aware the medImporter file is now massively bloated, and some further code cleanup can/should be done, but from an architectural perspective we've cut out some unnecessary problems and reduced coupling in the database handling.

### DIFF
--- a/src/medCore/database/medDataManager.cpp
+++ b/src/medCore/database/medDataManager.cpp
@@ -16,7 +16,7 @@
 #include <QDebug>
 
 #include <medAbstractDataFactory.h>
-#include <medImporter.h>
+#include <medDatabaseImporter.h>
 #include <medDatabaseController.h>
 #include <medDatabaseNonPersistentController.h>
 #include <medDatabaseExporter.h>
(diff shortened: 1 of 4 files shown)


fvichot:
I just noticed you renamed ```medAbstractImporter``` to ```medImporter```, when for consistency it should be ```medDatabaseImporter```.

rdebroiz:
Seems to be a good refactor of the DB. 

But i have some troubleshooting while loading some plugins (some methods of some class are not implemented anymore I guess)

Anyway, it doesn't fix any bug or improve something for a user point of view, so I think it would be safer to look for merge this one after the next release. 

;)

rdebroiz:
> But i have some troubleshooting while loading some plugins (some methods of some class are not implemented anymore I guess)

Forget what I said, It's just a matter of rebase on the master.

ocommowi:
Should this be in 3.0? Clear milestone if not

PR #211, opened 2014-12-18
Changing window/level to min/max

gpasquie:
We had some comments here that window/level was not really handy to manage intensity and that min and max could be more useful.
Here is a version of it.
You can still change window/level with the mouse but you can also set min and max with sliders in the toolbox.

Tell me what you think.
Guillaume

### DIFF
--- a/src/medCore/parameters/medDoubleParameter.cpp
+++ b/src/medCore/parameters/medDoubleParameter.cpp
@@ -29,11 +29,7 @@ class medDoubleSpinBox : public QDoubleSpinBox
 public:
     virtual QString textFromValue ( double value ) const
     {
-        std::ostringstream oss;
-        oss << std::setprecision(7) << value;
-
-        QString res = QString(oss.str().c_str());
-        return res;
+        return QString::number(value, 'g', 7);
     }
 };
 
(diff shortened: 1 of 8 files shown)


fvichot:
This is a behaviour changed, not a bugfix, not sure this should go in a minor release.

mikejbuckingham:
Looks good for me, but I'll ask for a second review just in case, as I'm new to Qt.

ocommowi:
We said on monday that this was a long wanted feature both in Rennes (several people asked us repeatedly for that) and in Sophia (Max and probably others). So that's why we've put it for 2.2.2. 

That being said I don't mind having it in 2.2.2 or 2.3. It's just that the next big release will probably be around the summer. So the question to ask is do we want to have it included soon or can we wait for june or even september ? That's your call :)

src/medCore/parameters/medDoubleParameter.cpp
```diff
@@ -20,14 +20,32 @@
...
+{
+public:
+    virtual QString textFromValue ( double value ) const
+    {
+        std::ostringstream oss;
+        oss << std::setprecision(7) << value;
+
+        QString res = QString(oss.str().c_str());
+        return res;
+    }
```
fvichot: What's wrong with QString::number(value, 'g', 7) ?

PR #212, opened 2014-12-18
Job status widget

ocommowi:
Just not to forget it, this is the branch implementing the job status widget. It implements everything based on the process refactoring branch so merge should be done there for now. The only thing that can most probably be perfected is the aspect of the widget, especially its resizing.


ocommowi:
This is based on #213 , it should be reviewed after it.

rdebroiz:
Shouldn't we rebase/merge it with process refactoring?

mikejbuckingham:
I'm also in favour of rebasing this on process refactoring, but let's keep this open for now so we don't forget.

PR #213, opened 2014-12-19
Process refactoring

gpasquie:
This PR replaces #188.
Branch is now on directly on medInria repo.


gpasquie:
I pushed some last minute modifications about the registration workspace. You should now be able to apply linear, displacement field and SVF transformation. The diffeomorphic demon plugin should also be able to estimate transformations.
Some stuff left to do though:
- how to identify the link between the estimation process and the process to apply the transformation
- change the way I retrieve the containers, use a static method provided by the process to apply the transfo instead.
- improve the way we handle the registration output (naming and avoid to generate a new data every time if possible)
- cleanup in itkRegistrationProcess / Factory
- We created a medAbstractUndoRedoProcess but it's currently not used. Instead, I added a QUndoStack and a QundoView in medAbstractApplyTransformationProcess. Maybe, we should rethink medAbstractApplyTransformationProcess API and integrate these 2 element in it.

RP is quite aware of what has been done so whoever who continues working on this, don't hesitate to communicate with him.

PR #217, opened 2015-01-07
Crash when empty study

gpasquie:
This PR should correct a crash that happened when you were hovering the thumbnail area after creating a new study


fvichot:
Works for me!

ocommowi:
This works for me, so we can merge I guess

Issue #218, opened 2015-01-08
command line options


gpasquie:
When launching medinria with the command line, the --view option enables to directly open an image. However, when you use this option to open several files at the same time, all the files are opened in the same container. It could be interesting to add an option to open the different files in different containers.

Issue #198, opened 2014-12-09
Cannot select multiple files loaded in medInria


bishesh:
I cannot select multiple files that are present on the left window pane under (Patient/Study/Serie). 
I usually have to visualize many files at the same time so I select all of them from the hard disk using "Browser" and then use the action "Load". 
After working with them when I want to "remove" these files, I have to remove the files individually since I can't select all of them together. I want to remove them to free up the RAM space and load other images for visualization.

mathildemerle:
This is solved

PR #220, opened 2015-01-08
Multiple select and removal in browser

mikejbuckingham:
As requested in issue #198 

For the time being, all we do here is enable selection of multiple rows, and provide the "Remove" functionality. In the future, it might be worth adding some extra options (such as view, but we need to decide how we're handling multiple-input views first (As mentioned in #218)

As we get cascade deletion of children when removing a parent, there is a situation where we're requesting th delete of an element that has already been removed. However, the DB code handles this quite gracefully, and I don't think it's the business of the viewer to intrude here. Would welcome feedback though (particularly as I plan to revisit the DB code in the near future)

### DIFF
--- a/src/medCore/gui/database/medDatabaseView.cpp
+++ b/src/medCore/gui/database/medDatabaseView.cpp
@@ -11,6 +11,8 @@
 
 =========================================================================*/
 
+#include <qnamespace.h>
+
 #include <medDatabaseView.h>
 #include <medDataManager.h>
 #include <medAbstractDatabaseItem.h>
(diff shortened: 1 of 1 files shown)


src/medCore/gui/database/medDatabaseView.cpp
```diff
@@ -11,6 +11,8 @@
 
 =========================================================================*/
 
+#include <qnamespace.h>
```
fvichot: This looks like an internal Qt header, I don't think you should include it here.

gpasquie:
It seems to work well for multiple deletion, good stuff. 
However, I think there are some problems.
First, as you updated the contextual menu when you have multiple items selected, you'll need to update the "actions" toolbox on the left panel, otherwise the user can still have the feeling he can view or export multiple items.
Moreover, and it's a bit more complicated I think, you'll have to manage moving several items in the db...We can imagine being able to move multiple series at the same time but what do you do when you have a serie and a study selected at the same time? Should we disable drag and drop?
Or maybe we should allow  multiple selection only for the same types of items but I'm not sure this is easily doable...

rdebroiz:
>Or maybe we should allow multiple selection only for the same types of items but I'm not sure this is easily doable...

+1

ocommowi:
Is this solved or not? Should it be in 3.0 ?

Issue #215, opened 2014-12-30
Depth peeling does not work anymore


fvichot:
When depth peeling is activated, it does nothing. Also, it should be moved out of the mesh navigator and moved into the vtkViewNavigator, as it applies to fibers as well, and in general anything rendered with polygons.

ocommowi:
Well, I just tested and it does work for me on head data meshes in medinria data. At least, it does something good on those. Do you have an example data that used to work and works differently ?

fvichot:
Apparently, it's a new driver problem... so far, it doesn't work on nvidia drivers on Linux, either on F16 or F19, but works on intel on linux or mac, with or without software rendering. If anyone has a Nvidia on Mac, I'd be curious to know if it works.

fvichot:
Just tested on Macleod, which has an ATI GPU, and it doesn't work either...

mikejbuckingham:
So the "correct" fix for this is to do depth sorting, but this introduces a number of problems in our pipeline.

First, vtk's depth-sorting can only be performed on a single mesh, and in order to get the correct behaviour we'll have to create a unified mesh from all layers, then perform depth sorting on that mesh before rendering. This makes handling opacity problematic, as we're now operating on a single mesh (with a single property) rather than each layer having its own properties.

Secondly, because the depth-sorting is obviously subject to camera angles, the algorithm is run each frame. This is prohibitively expensive in terms of performance.

As this appears to stem from a driver issue rather than a problem in our code specifically, I'm recommending we don't try to fix it on our side, but rather wait for the drivers to be fixed.

mikejbuckingham:
Extra info : Just tested on a Mac with an nVidia GPU, works fine.

mikejbuckingham:
So the vtk code has this fantastic little snippet :
else if(isATI)
{
    this->DepthPeelingIsSupported = 0;
}

So I guess we'll not be providing depthpeeling on ATI cards anytime soon! I'll add a tooltip to state that depth peeling doesn't work on all architectures.

Still no idea why nvidia doesn't work yet.

fvichot:
[...]

PR #223, opened 2015-01-12
Fixes depth peeling not working on all platforms

fvichot:
Since this is a pretty trivial change, and actually put things back to how they were before the view refactoring, I'm thinking we add it to the release ? This also updates the release notes.


ocommowi:
If it does solve the problem, I'm for putting it in the new release. Codewise, it is ok for me (not quite sure though why this is solving something but I'm a vtk newbie). Can anyone test it on various platforms ?

fvichot:
Read the issue #215 ;)

rdebroiz:
Fix the issue for me. 

OS: Ubunutu 14.10
GPU: Nvidia  
dirivers : proprietary

PR #224, opened 2015-01-14
Fix for offscreen-rendering thumbnail crash

### LINKED ISSUE #201: Crash during thumbnail generation of images on linux machine with ATI GPU.
On machine that have (at least) ATI GPU, with (at least) ubuntu, medInria crash during the thumbnail generation of images.

Here the call stack associated:
```
0    XQueryExtension    /usr/lib/x86_64-linux-gnu/libX11.so.6        
1    XInitExtension    /usr/lib/x86_64-linux-gnu/libX11.so.6       
2    ??    /usr/lib/x86_64-linux-gnu/mesa/libGL.so.1        
3    ??    /usr/lib/x86_64-linux-gnu/mesa/libGL.so.1      
4    ??    /usr/lib/x86_64-linux-gnu/dri/i965_dri.so       
5    ??    /usr/lib/x86_64-linux-gnu/dri/i965_dri.so        
6    ??    /usr/lib/x86_64-linux-gnu/dri/i965_dri.so       
7    ??    /usr/lib/x86_64-linux-gnu/dri/i965_dri.so        
8    ??    /usr/lib/x86_64-linux-gnu/dri/i965_dri.so        
9    glXMakeCurrentReadSGI    /usr/lib/x86_64-linux-gnu/mesa/libGL.so.1        
10    vtkXOpenGLRenderWindow::MakeCurrent() 
11    vtkXOpenGLRenderWindow::WindowInitialize()   
12    vtkXOpenGLRenderWindow::Start()    
13    vtkRenderWindow::DoStereoRender()          
14    vtkRenderWindow::DoFDRender()   
15    vtkRenderWindow::DoAARender()    
16    vtkRenderWindow::Render()   
17    vtkXOpenGLRenderWindow::Render()   
18    vtkImageView::Render    vtkImageView.cxx  
19    medVtkViewNavigator::changeOrientation    medVtkViewNavigator.cpp
```
I strongly suspect that it is again an offscreen renderring problem. Visualizing an image without generating a thumbnail works like a charm.

Note that i don't have this issue with an Nvidia GPU and I didn't test with Intel GPU.

mikejbuckingham:
This PR fixes #201

Instead of using offscreen rendering, we're creating a new window and hiding it behind the main application, and then rendering normally to that window. From there, we take the thumbnail.

This is, in short, a total hack. It is also unfortunately the best solution we could come up with, considering how offscreen rendering dies horribly in the mesa drivers.

The performance is adversely affected here - because of x11's fanatical wish to animate the creation of new windows, we have to wait for it to finish before we can use the window. Because we manipulate the window twice, we have to wait for x11 twice. Unsurprisingly, this wrecks our performance.

I would suggest that we detect which configurations are susceptible to the bug, and only perform this hack in those cases, and preserve the standard offscreen rendering for the other, stable configurations. 

What does everybody else think?


rdebroiz:
Sounds ugly but it seems to actually works like a charm. Well done.

fvichot:
I added some code to check the GPU vendor, and disable offscreen rendering in the combinations known to crash, so far ATI+Intel on Linux, and Intel on Linux. I've only tested on NVidia Linux that the detection code works properly, so please test if you have other combinations of OS and GPU vendor.

mikejbuckingham:
Just tested Florian's changes, works fine for me. On our respective machines now, Florian's nvidia does proper offscreen rendering, while my Intel does the hidden-window-render hack.

fvichot:
@ocommowi can you review on mac and merge if OK ?

ocommowi:
I'm missing a library link on Mac. Could you add an explicit link to the appropriate lib ?

Undefined symbols for architecture x86_64:
  "_glGetString", referenced from:
      med::gpuModel() in medGlobalDefs.cpp.o
ld: symbol(s) not found for architecture x86_64

fvichot:
hm that should be taken care of by ```set(QT_USE_QTOPENGL on)```

ocommowi:
Well, apparently not. It's probably that you need somewhere to explicitly link again QtOpengl or something like that (mac linking is not tolerant with respect to that kind of thing)

ocommowi:
Adding an explicit link to ${OPENGL_gl_LIBRARY} in medCore CMakeLists does the trick on Mac. Does that cmake variable exist also on windows and linux ?

fvichot:
Should work now

ocommowi:
It works great. Merging then

Issue #227, opened 2015-01-21
DB schema issues


fvichot:
The DB schema needs to be updated. Notably:
- the **image** table is currently useless, as we always store series as a single file (so as a single row in image). Also, there is already a _path_ field in **series**.
- **patient** and **study** tables have a _thumbnail_ field which is unused
- Currently, on the disk we have :

```
<db_directory>/
           <study_UUID>/
                     <data files named with series UUID>
                     <a folder of thumbnail(s)>
```

but we don't move series out of the study folder when they get moved to an different study. So it would make a lot more sense to just have:

```
<db_directory>/
           <series_UUID>/
                     data.<proper_extension>
                     thumbnail.png
```

and then just have the patient/study info being stored in the DB.
- Also, we have a ton of metadata fields in the **series** table, but we can't really dynamically add new ones without modifying the DB. I'm thinking we should instead have a **metadata** table with the following schema: 

```
CREATE TABLE metadata (id INTEGER PRIMARY KEY, name TEXT, value TEXT, ownerId INTEGER);
```

and change the **series** table to : 

```
CREATE TABLE series ( id INTEGER PRIMARY KEY, studyId INTEGER, name TEXT,
                      path TEXT, seriesUuid TEXT);
```

That way, all the optional metadata goes as name/value pairs in **metadata**. For display, since we could no longer have designated columns for all the fields I just removed, I'd advise adding a right pane in the browser view, with a list of all the name/value metadata pairs, that you could modify, and obviously allow for search on these, probably using a syntax like `<name>:<value>` in the search field instead of adding search constraints via a menu like we do know. Of course metadata on patients/study could be done the same, by changing the **patient** and **study** tables to these:

```
CREATE TABLE patient ( id INTEGER PRIMARY KEY, name TEXT, patientUuid TEXT);
CREATE TABLE study ( id INTEGER PRIMARY KEY, patientId INTEGER, name TEXT, studyUuid TEXT);
```

You'll notice I've kept the UUIDs for studies and patients, but I'm not entirely sure they're still necessary. If you move patient and study metadata into the **metadata** table, you'll need to update the **metadata** schema by adding a _ownerType_ field to select whether the _ownerId_ should be looked up in the **patient**/**study**/**series** tables. Probably just an `INTEGER` with 0=**patient**, 1=**study** etc.

fvichot:
[...]

Issue #225, opened 2015-01-19
Thumbnail access


rdebroiz:
For now there are two ways to access the thumbnail of a data:

``` c++
QImage& medAbstractData::thumbnail() 
//it should return QImage or QImage const&  IMO by the way

QPixmap medDataManager::thumbnail(const medDataIndex& index); 
// the method should be const IMO by the way
```

The problem is that the first one is not an accessor, it actually tries to generate a thumbnail if the `d->thumbnail` attribute is an empty image. Since there is no setter to this attribute it's always the case...

IMO we should fuse this method with `void medAbstractData::generateThumbnail()` 
which is private and make it friends with the class that are actually supposed to do that, which are `medDatabaseNonPersistentImporter` and `medAbstractDatabaseImporter` apparently. So the medDataManager will be the only access point to the thumbnails.

mikejbuckingham:
I agree in principal with the idea of merging the methods, but I'm not sure I see the reason for friendship - why not simply publicise the generateThumbnail method?

Agree on the const-correctness, and probably renaming stuff to be more explicit (getThumbnail vs thumbnail)

rdebroiz:
> - why not simply publicise the generateThumbnail method?

Because it can crashes if it is improperly called. For example it crash if you call it from a non smart pointed data (but maybe the real probleme is that we can manipulate data that have not be smart pointed at least once ;-)). Plus I can't find other cases than import  where it can be usefull.

fvichot:
Actually, apart from the const-correctness which I agree should be just QImage, and probably even QPixmap for consistency with the other method, the rest I don't agree. ```medAbstractData::thumbnail()``` and ```medDataManager::thumbnail()``` have different roles. The first one is to get the thumbnail from a loaded data object, the second is for loaded a thumbnail from a DB item, without loading it.

fvichot:
Grrr I was going to change the signature from ```QImage&``` to ```QPixmap``` in PR #224 , but obviously since it's a method from dtkAbstractData, I can't change the return type without also changing the name. So do I leave it like that with a comment to change it when we get rid of the dtkAbstract* stuffs, or should I change it to ```getThumbnail()``` or something and do it now?

rdebroiz:
Ok, since only images aka series can have thumbnails, and because thumbnails have sense only in the database i can't see why ```medAbstractData::thumbnail()``` should exists.

[...]

PR #229, opened 2015-01-23
change thumbnail() to generateThumbnail() and DB cleanups

fvichot:
Implement the changes discussed in #225, as well as some DB cleanups based on #227 , notably I removed all notion of the "image" table in the DB.


rdebroiz:
When you saved an image in the persistent DB the thumbnail is not saved, I suspect that you are trying to write the QImage in a non existing folder.

fvichot:
**Make sure to backup your "db" file before you test this, as it will modify the schema of the series table, and drop the image table.**

So this as far as I got:
 - I've removed all traces of the "image" table
 - I've migrated the isIndexed field from image to series
 - the path field in series table is now used for the image(s) path(s). If more than one path, they are separated with ';'. This way we should retain the ability to index multi-files data items, like DICOM (not an issue in case of import as we convert it to a single file, a MHA in the case of DICOM).

If someone continues on this, one of the tricky bits is not breaking the whole DB, have a look at how I've implemented ```medDatabaseController::updateFromNoVersionToVersion1()```. Also, you'll notice I've started versioning the DB schema using ```PRAGMA user_version``` which seems like a good idea.
 This PR should also be merged with PR #207 at some point too, and as such it would provide a nicer base to refactor the importer during the upcoming data refactoring.

PR #231, opened 2015-01-27
Fix registration for unsigned short images

gpasquie:
Reenabling registration for unsigned short images (like I1.nhdr) that was not possible in medInria2.2.2


mikejbuckingham:
Seems good to me, very straightforward.

ocommowi:
Fine for me as well. Merging

PR #234, opened 2015-02-06
Add tests on NaN values to avoid crash during window-leveling

rdebroiz:
The invert intensity image filter is broken and make always images with really high values (1.798^308)

If you change the WL of the resulting image, it try to apply a window or a level with NaN and it crashes. Here is a patch to fix the crashes caused by the NaN.

I'll fix the invert intensity image filter soon.


mikejbuckingham:
Fine for me - both MSVC and GCC are fine with this, but not all compilers are. Something we need to bear in mind before we jump to C++11 (where we can use std::isnan)

PR #235, opened 2015-02-06
Deactivate invert image filter for double and float images.

rdebroiz:
Invert image filter doesn't work with double and float images


mikejbuckingham:
Looks fine to me - will merge if no objections at the meeting tomorrow.

@rdebroiz , one point: Could you try to avoid cosmetic changes like the whitespace? I know it shouldn't be there in the first place, and in this case it's not too big of a deal, but it clutters the diffs a bit so it is slightly harder to see where the real changes take place.

PR #237, opened 2015-02-06
Set default Local of Qt as the C one

rdebroiz:
Otherwise it can be very annoying for the edition of some widgets such as QDoubleSpinbox on some systems:

The SpinBox allow to edit only with coma as floating separator but allow to only display dot.

After this patch, dot is always used.


ocommowi:
Fair enough to me. If no one is against it, I'll merge

mikejbuckingham:
Works for me!

PR #238, opened 2015-02-12
Add a colormap for images that are label maps

ocommowi:
A tentative binary map LUT to better see label images. Tell me what you think.

PS : by the way, this made us realize (me and RP) the LUT/transfer functions system in medinria is quite a bit crappy. @papadop : any chance your branch on that could be revived / finished soon ?

### DIFF
--- a/src-plugins/itkDataImage/interactors/medVtkViewItkDataImageInteractor.cpp
+++ b/src-plugins/itkDataImage/interactors/medVtkViewItkDataImageInteractor.cpp
@@ -223,7 +223,7 @@ void medVtkViewItkDataImageInteractor::initParameters(medAbstractImageData* data
                                      << "Asymmetry" << "P-Value" << "Red Black Alpha"
                                      << "Green Black Alpha" << "Blue Black Alpha"
                                      << "Muscles & Bones" << "Bones" << "Red Vessels"
-                                     << "Cardiac" << "Gray Rainbow" << "Stern" << "Black Body";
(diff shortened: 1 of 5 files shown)


msermesant:
@LoicCadour : how does it compare to the discrete LUT feature you developed in MUSIC?

LoicCadour:
Actually I created a mode in the medClutEditor to switch between linear and discrete mode. I didn't create a color map, I just modify the current LUT according to the mode.

ocommowi:
Well, here it's a LUT with a different color for each label.Since it is a LUT, it is adapted depending on the min max values as the other LUTs. It has advantages and disadvantages, let me know what you think of it.

LoicCadour:
In MUSIC, we needed this discrete mode to clearly visualize few (in most cases 2 or 3)  groups of pixels according to their intensity values. For instance, pixels whose intensities belong to [0;100] are displayed in red, [101;200] in green etc. And this is very easily done by placing vertices on the histogram.
I think your colormap is more useful for label images, in our case we are playing with CT/MR images.

mikejbuckingham:
@LoicCadour @ocommowi 

What's the consensus here? Can we merge this, close it, or is more discussion needed?

LoicCadour:
I quickly tested this branch (with MUSIC) it was working fine. The things I've done with the ClutEditor are not really related, so I've nothing else to add.

mikejbuckingham:
OK - will merge after the meeting, unless there are objections.

PR #241, opened 2015-03-02
Persistency: Save images with their interactors/navigators

NicolasSchnitzler:
Allows to save and reload a scene (active workspace, displayed images, interactors and navigators) in/from a folder.
Absolute paths are used to save/read the files from their location, but are saved as relative paths so the folder containing the data can be moved.
Does not handle views layout.


rdebroiz:
It seems that you have twice the same changes in your 20 firsts commit. ie: 
https://github.com/NicolasSchnitzler/medInria-public/commit/16c6cac2ebe9503d40a732908f705f790e153d32 and  https://github.com/NicolasSchnitzler/medInria-public/commit/254bcb2fc2c7a618bdaa8d37a24a1bee531ed4f6 are the same.

I guess something went bad during a rebase or a merge...

rdebroiz:
I just made a new PR without the duplicates commits, check this out here: https://github.com/medInria/medInria-public/pull/243 

Very nice work, a lot has been done. 

- When you rewrite image data, you use the gipl format, it end in tones of message on the cerr output because the DCMTK reader tries to read it. I Guess a better choice would be to use the same as we use in the database (.mha if i'm not wrong).

- It doesn't work for vtk files (meshes, etc.)

- the modal window to load / save scenes saves the path choose in the settings and overwrites the one used to open an image,  which is very annoying.

- We rewrite the file of each data being visualized, I assumed that the goal of it, is to be able to send the serialised scene to someone else on another machine, which is a nice feature. But maybe there could be a sort of mode to change this behaviour if one want to save a  scene on this machine for later, without duplicating all the data. I understand that's not so easy because data can come from persistent DB non persistent DB, output of process, etc.

mikejbuckingham:
Can we close this, and move over to the new PR for the review?

PR #243, opened 2015-03-04
Scene saving, no duplicates

rdebroiz:
Same as https://github.com/medInria/medInria-public/pull/241, without duplicated commits.


NicolasSchnitzler:
following the discussion on #241
>    When you rewrite image data, you use the gipl format, it end in tones of message on the cerr     output because the DCMTK reader tries to read it. I Guess a better choice would be to use the   same as we use in the database (.mha if i'm not wrong).

>    It doesn't work for vtk files (meshes, etc.)

>    the modal window to load / save scenes saves the path choose in the settings and overwrites     > the one used to open an image, which is very annoying.

>    We rewrite the file of each data being visualized, I assumed that the goal of it, is to be able to     > send the serialised scene to someone else on another machine, which is a nice feature. But          > maybe there could be a sort of mode to change this behaviour if one want to save a scene on this  > machine for later, without duplicating all the data. I understand that's not so easy because data     > can come from persistent DB non persistent DB, output of process, etc.

For point 1, I can easily default it to mha, so far the policy was to use the first file extension that works. To prevent looking through all the writers each time, would it be correct to add the file extension to the metadata?

For point 2, it appears that the problem is currently the vtkDataMeshWriter::canWrite method that is returning false.
the code of this method is shown below: 
```C++
bool vtkDataMeshWriter::canWrite(const QString& /*path*/)
{
    if ( ! this->data())
        return false;

    return dynamic_cast<vtkMetaDataSet*>((vtkObject*)(this->data()->data()));
}
```
the first condition is always false, so the function returns false, and the writer is never used. Furthermore, even if this first condition was verified, the dynamic_cast would fail, given vtkDataMesh does not inherits of vtkMetaDataSet.
So, 
- is this method still useful, or just legacy code ? (export feature from the db does not use it, the      write method already performs the same checks, commenting the calls to this method in medDataReaderWriter allows to save meshes)
- if it is performs a check, how should i use it?

For point 3, it will be fixed easily 

For point 4, i guess it needs to be discussed further (what do we want to save precisely? what do we want to do if a file was moved/deleted/renamed/is on another machine?)

LoicCadour:
Really nice work! 
I tried it with I1 and I2 in a split view. It kept the organisation of the views, the LUTs, the opacities, etc.
However, I couldn't get the right windowing though the Min/Max intensity values were set in the xml file.

Issue #233, opened 2015-01-30
Data refactoring


fvichot:
So once the process refactoring is done (#213), I would advise tackling the DB (#227) and importers (#207). From there, here are my suggestions:

**Importer**
 - After PR #207 , there should be just one ```medDatabaseImporter``` which gets provided a controller on creation so that it can interact with the either the persistent or none persistent DB transparently. 
 - There is still work to be done so that the role of the importer and reader are not mixed, currently the importer knows a bit too much about DICOM, and that knowledge should be moved back into the reader. After discussion with @mikejbuckingham , the suggested strategy was to have a ```bool canRead(QString path, QString& volumeId)``` method in the readers, which as well as telling the importer whether this reader can handle this file, tells it if it part of a bigger cluster of files. That way the ```volumeId``` currently being computed in the importer based on the extracted metadata of each file can be moved into the DCMTK reader, and we keep the knowledge of DICOM in the reader. This should massively simplify ```importFile()```.

**Reader**
- Then, it would be time to consider retiring the itkDMCTKImageReader, and replacing it with the GDCM based reader from DICOM (see http://www.itk.org/Wiki/ITK/Examples/DICOM/ResampleDICOM for an example).

**Data**
 - In prevision of DTK 1.0, it might be a good opportunity to drop the inheritance from ```DtkAbstractData``` since it will be deprecated.
 - The above means that we will lose the reference counting currently done in DTK, so if they don't provided a way to keep it, that means reimplementing a medSmartPointer, and a general ```medObject``` class in which to keep the refCount. When doing this, keep #226 in mind.
 - @rdebroiz had a proof of concept somewhere about simplifying the image classes, so that we just store a regular pointer instead of having to template cast everything, probably a good idea to have a look and integrate if viable.

PR #244, opened 2015-03-06
Itk data improvements

papadop:
This is a cleanup of itkDataImage. Several things where done:
- Put the private types in their own header file.
- Suppressed number of includes that were no longer used.
- Simplify the mecanism that uses two different private types for scalar and vectorial type. No longer use a ugly specialization. I believe the stuff is now simpler.

Future work: Split further itkDataPrivateTypes.h ???

### DIFF
--- a/app/medInria/medApplication.cpp
+++ b/app/medInria/medApplication.cpp
@@ -170,7 +170,7 @@ void medApplication::initialize()
     viewerWSpaceFactory->registerWorkspace<medSegmentationWorkspace>();
 
     //Register settingsWidgets
-    medSettingsWidgetFactory * settingsWidgetFactory = medSettingsWidgetFactory::instance();
+    medSettingsWidgetFactory* settingsWidgetFactory = medSettingsWidgetFactory::instance();
     settingsWidgetFactory->registerSettingsWidget<medStartupSettingsWidget>();
     settingsWidgetFactory->registerSettingsWidget<medDatabaseSettingsWidget>();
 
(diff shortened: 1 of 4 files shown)


rdebroiz:
I think the future work will be to not use template at all for `itkDataImage` types, and have only one `medItkImageDataType` for any data that encapsulate an `itk::Image`. 

This is possible by encapsulating the image with a pointer of `itk::Object` instead of `itk::Image<PixelType, Dimension>`. 

When you want to use it by casting the void* poiter returned by `void* medAbstractData::data()` you can dynamic_cast it in the right `itk::Image<PixelType, Dimension>` using the `unsigned int medAbstractImageData::dimension()`, `PixelType medAbstractImageData::pixelType()`, etc. in the medAbstractImageData  API.

We already discuss about it with Flo before, more info: https://github.com/medInria/medInria-public/issues/233

papadop:
I do not see anything in #233 related to this....
In my view, this is to be discussed...
What does it try to solve ???
This goes against more type safety...
Nevertheless, the proposed patch just corrects some wrong C++ code... It has nothing to do with the future.

rdebroiz:
Last part of #233:

> @rdebroiz had a proof of concept somewhere about simplifying the image classes, so that we just store a regular pointer instead of having to template cast everything, probably a good idea to have a look and integrate if viable.

The idea is basically to don't have to write a plugin each time we want to support a new PixelType/number of Dimension/Number of component (most important) in itk::Image

mikejbuckingham:
I'm going to merge this by Monday morning, if no objections. While I can see that future work might not actually benefit enormously from this, this code is ready to go today and it costs nothing to integrate, while leaving it here doesn't help anyone.

Please let me know if there are issues with merging this before Monday noon.

PR #246, opened 2015-04-02
Bringing back data converters

ocommowi:
Converters were allowed by DTK abstract data to define conversions between data types. For some old data types that we still use for demos, I had to bring them back in medAbstractData. It can be something useful for future data types people will handle.

This also solves a bug on mac where the plugin generator was generating a registration plugin when we were asking for a filtering plugin

### DIFF
--- a/app/medPluginGenerator/medPluginGenerator.h
+++ b/app/medPluginGenerator/medPluginGenerator.h
@@ -35,7 +35,7 @@ enum pluginType
     IMAGE_NAVIGATOR,
     EXTRA_NAVIGATOR,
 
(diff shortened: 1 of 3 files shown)


mikejbuckingham:
At the risk of sounding stupid, I'm not sure what this does? I can understand the code, but is convert actually called by anything? This code doesn't look used to me.

rdebroiz:
guess it's used in one of our private plugins ;)

mikejbuckingham:
Yeah, I figured that would be the case, but didn't want to say "OK" without checking. Looks good for me then.

PR #249, opened 2015-04-27
Change format of scalar bar text to scientific

rdebroiz:
This is the most elegant way IMO to always have a good display of any value (either very small or very big).
It also fix the Nan bug on windows (CRASH) (DrawNanValue is coming with Vtk 6 to correctly display Nan values on the scalar bar.


LoicCadour:
It indeed resolves the Nan pb on Windows. Well done!

ocommowi:
Well, tested it and it works fine. Not the greatest but if it solves a crash, good for me. Merging, may be keep in mind this story of vtk 6 if it can bring a new solution to this

PR #250, opened 2015-06-12
Switch to VTK6

NicolasSchnitzler:
Now working with CT scans and 4D images.
there is still the spherical harmonics bug, and further testing should be made for software rendering (working for me, but it seems that MESA does not support floating point textures).

The crash when switching back to VR view is now fixed, but i am not sure what we want to do:
-delete the 3D view each time we switch back to 2D which implies reloading fully the 3D when we go back to it.
-keep the 3D view, which imply not releasing resources for 3D view.
In my understanding the crash came from the fact we are currently mixing these two policies.

(The concerned classes there are vtkImageView, vtkImageView2D and vtkImageView3D in the folder src/medVtkInria/vtkImageView.
The functions named UninstallPipeline (l 510, 1664 and 485 respectively) the crah occurs when the vtk renderer tries to access opacity table, that has already been released)


ocommowi:
Well, well, well, here comes the annoying tester :) First of all, great work for all the bugs you corrected. Here are the ones I see remaining (I know you already know some of them:

- window leveling on the first layer is very slow. Take any image large enough, and it will lag. That was not the case with vtk 5 and it's also working fine if the image you're window-leveling on the second layer
- window leveling is not linked between 2d and 3d MSR modes (but link with VR does work)
- viewing an ODF or tensor image alone is working fine, viewing it as a second layer over another image makes the app crash. Testing on tensor images (with the DTI cube example of medinria) makes me believe the input algorithm is returning a null w_ext pointer for some reason I don't get. I don't know if it's the same for ODFs
- when viewing a tensor image, switching sample to 1 makes the image empty
- viewing 4D images works, but timeline changes are not linked between 2d and 3d modes
- finally, talking about slowliness again, try to view twice the same 4D image, link them and play them fast in the timeline. You'll see that they are not rendering exactly together and it gets worse with time. That was not the case before.

All in all that's a lot of bugs and regressions. I would love to see the major ones (tensor crashes and wrong behaviors, slowliness) solved before merging

NicolasSchnitzler:
outdated, closing

PR #251, opened 2015-06-16
Fix for broken camera

mikejbuckingham:
Previous behaviour :

Open mesh in VR mode
Hold CTRL and leftmouse, and move mouse horizontally, to rotate camera around the Z-axis.
Hold leftmouse and move mouse horizontally - doesn't correctly rotate around the Y-axis anymore.

New behaviour:

As above, except rotation around the Y-axis is no longer broken by rotating around the Z-axis.

The problem came from the fact that z-rotation is handled by vtk directly, while x and y rotation is handeld by our own code, that relies upon certain variables being updated. vtk obviously doesn't update those variables, as it doesn't know about them, so we don't correctly handle the change in camera y-axis after a z-rotation.


ocommowi:
Works well, simple enough. Merging

PR #253, opened 2015-06-30
Try to read temporal resolution from DICOM tag "Repetition Time"

msermesant:
fix for wrong duration of cardiac sequences by setting the right resolution time. some sequences stil have problems, maybe due to parallel imaging.

### DIFF
--- a/src/medImageIO/itkDCMTKImageIO.cpp
+++ b/src/medImageIO/itkDCMTKImageIO.cpp
@@ -505,7 +505,23 @@ void DCMTKImageIO::DetermineSpacing()
     }
 
     if (this->GetNumberOfDimensions()==4)
-        m_Spacing[3] = 1.0;
+    {
+        // try to use the Repetition Time as an estimate of temporal resolution (should work for cardiac)
+        const StringVectorType &temporalSpacing = this->GetMetaDataValueVectorString ("(0018,0080)");
+        if (temporalSpacing.size())
+        {
+            std::string temporalSpacingStr = temporalSpacing[0];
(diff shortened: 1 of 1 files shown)


src/medImageIO/itkDCMTKImageIO.cpp
```diff
@@ -505,7 +505,23 @@ void DCMTKImageIO::DetermineSpacing()
...
+        // try to use the Repetition Time as an estimate of temporal resolution (should work for cardiac)
+        const StringVectorType &temporalSpacing = this->GetMetaDataValueVectorString ("(0018,0080)");
+        if (temporalSpacing.size())
+        {
+            std::string temporalSpacingStr = temporalSpacing[0];
+            std::istringstream is_stream( temporalSpacingStr );
+            if (!(is_stream >> m_Spacing[3]))
+                itkWarningMacro ( << "Cannot convert string to double: " << temporalSpacingStr.c_str() << std::endl);
+            if(m_Spacing[3] != 0.0)
+                m_Spacing[3] *= 0.001;
```
mikejbuckingham: Can you put a comment explaining why you do this? I understand the rest, but I'm not really following the reason to take a 1/1000th? Is this arbitrary or following a standard?

msermesant: sorry, I forgot to comment this line: the DICOM tag is in milliseconds, and we work in seconds, thus  this scaling.

mikejbuckingham:
Good for me, will merge tomorrow if no objections.

PR #255, opened 2015-06-30
Categorise workspaces and clean up home page

msermesant:
I added a category to workspaces so that they can be organised on the home page. I also created columns for different categories on the home page.

Workspaces in private repos will have to be updated accordingly.


mikejbuckingham:
Good for me, will merge tomorrow if no objections.

PR #258, opened 2015-07-21
Switch to cmake 3

ocommowi:


mikejbuckingham:
As I recall, this has an issue with MacOS 10.6 or earlier?

If so, I will delay the merge until after 2.2.3 - let me know please.

ocommowi:
Not to be in 2.2.3, rather for 2.3.0 and for when we reset the jenkins slaves (we'll put cmake 3 on those slaves I guess).

rdebroiz:
It's now redundant with https://github.com/medInria/medInria-public/pull/265.
@ocommowi  do you care if I close this one ?

ocommowi:
Does that answer your question ? ;)

PR #259, opened 2015-07-23
Making some methods public (easy to review)

LoicCadour:
Teeny tiny changes to access important methods.

### DIFF
--- a/src/medCore/process/medAbstractProcess.h
+++ b/src/medCore/process/medAbstractProcess.h
@@ -32,6 +32,8 @@ public:
     medAbstractProcess( medAbstractProcess * parent = NULL );
     virtual ~medAbstractProcess();
 
+    using dtkAbstractProcess::setInput;
+
 public slots:
     virtual medAbstractData *output() = 0;
     virtual int update () = 0;
@@ -41,7 +43,6 @@ private:
     using dtkAbstractProcess::read;
     using dtkAbstractProcess::write;
     using dtkAbstractProcess::setParameter;
-    using dtkAbstractProcess::setInput;
     using dtkAbstractProcess::setData;
     //TODO rename our output methode
     //using dtkAbstractProcess::output;
--- a/src/medCore/views/medAbstractView.h
+++ b/src/medCore/views/medAbstractView.h
@@ -61,6 +61,7 @@ public:
     virtual QWidget* toolBarWidget();
 
     virtual QList<medAbstractNavigator*> navigators();
+    virtual QList<medAbstractNavigator*> extraNavigators();
     virtual QList<medAbstractInteractor*> interactors();
 
     medDoubleParameter* zoomParameter();
@@ -96,7 +97,6 @@ protected:
     virtual medAbstractViewInteractor* primaryInteractor();
     virtual QList<medAbstractInteractor*> extraInteractors();
     virtual medAbstractViewNavigator* primaryNavigator();
-    virtual QList<medAbstractNavigator*> extraNavigators();
 
     virtual bool initialiseInteractors(medAbstractData* data);
     virtual bool initialiseNavigators();


src/medCore/process/medAbstractProcess.h
```diff
@@ -32,6 +32,8 @@ class MEDCORE_EXPORT medAbstractProcess : public dtkAbstractProcess
     medAbstractProcess( medAbstractProcess * parent = NULL );
     virtual ~medAbstractProcess();
 
+    using dtkAbstractProcess::setInput;
```
ocommowi: I'm not really happy about that one. Our goal is here not to use the methods from dtk that may not be defined in some plugins. Rather, input methods are defined in the processes themselves, allowing more precise setInput parameters (medAbstractImageData or meshes, ...). Making it public may also create ugly warnings of hidden overloaded methods. And now public plugins don't use this setInput methods so you would expect it to do something but in fact it doesn't...

To summarize, I would prefer keeping it private. Don't you have ways to modify your processes and pipelines to keep that one private (it's just probably a story of method renaming) ?

LoicCadour: My pb is actually that in my plugins I use the dtkAbstractProcessFactory to reach my processes (for instance if I am in a plugin A and I want a process from a plugin B). 
I end up playing with a medAbstractProcess (I dont use dtkAbstractProcesse, I cant remember why), but I can't set his input.

src/medCore/views/medAbstractView.h
```diff
@@ -61,6 +61,7 @@ class MEDCORE_EXPORT medAbstractView: public dtkAbstractView
     virtual QWidget* toolBarWidget();
 
     virtual QList<medAbstractNavigator*> navigators();
+    virtual QList<medAbstractNavigator*> extraNavigators();
     virtual QList<medAbstractInteractor*> interactors();
```
rdebroiz: Why did you need to have extra navigators specifically? isn't navigators enough ? If you really need it, I would be more in favour to put all the 'primary /extra navigators / interactors' accessor public for consistency.

src/medCore/process/medAbstractProcess.h
rdebroiz: I see what you did here, but i'm not even sure that it will fix your probleme (you have no guarantee that your process use a setInput methode taking a type deriving from dtkAbstractData*) so maybe you will be to call the methode but it will do nothing.

Your probleme should be fix when the process layer will be design, you wil bel then able to use something like:

```cpp
medAbstractWhateverKindOfProcess *p = medProcess::whateverKindOfProcess::pluginFactory().create("whateverKindOfProcess_id");
```
where `medAbstractWhateverKindOfProcess` will expose the setter / accessor that you need to set the inputs or get the outputs for this specific kind of process.

LoicCadour: But how could the object know `medAbstractWhateverKindOfProcess` since it belongs to another plugin?

rdebroiz: medAbstractWhateverKindOfProcess do not belong to another plugins, it's defined in the process layer which comes from the libs ships by medInria (medCore, medProcess, medGui etc..)

it could be for instance medAbstractSubtractImageProcess (see the dev branch in the medInria-public repo) which expose: 
```cpp
void setInput1(medAbstractImageData *data);
void setInput2(medAbstractImageData *data);
medAbstractImageData * output() const;
```

LoicCadour: OK but my pb is that I create a process in a plugin. And in a pipeline (which belongs to another plugin), I want to use it.
Here is a clear example of what we are doing right now.

I am in my pipeline (pluginA) and I want to use a thresholding process defined in pluginB. This is how I do it:
```cpp
(pipeline.cpp)

dtkSmartPointer<medAbstractProcess> thresholdProcess = 
      dtkAbstractProcessFactory::instance()->createSmartPointer("itkThresholdingProcess");
thresholdProcess->setInput(maskedMRI);
thresholdProcess->setParameter(thresholds[i], 0); //thresholdValue
...
```
So what is needed is just a medAbstractProcess with accesible *setInput, setParameter*

rdebroiz: Yes I get your point. 

I'm just saying that it's not possible for now (You would be lucky that the implementation of itkThresholdingProcess uses a setInput(medAbstractData *data).

But it will be possible when the process layer will be present in the core of medInria, you will have then a medAbstractThresholdImageProcess or something like that exposing the methods that you need to setup the implementation of your process.

rdebroiz: What I mean is that the method dtkAbstractProcess::setInput is not used by medAbstractProcess that's why it redefined as private. It can be redefined as public  with a covariant type of dtkAbstractData* as argument in an implementation of medAbstractProcess (maybe it is in your particular case) but it's just a matter of luck.

src/medCore/views/medAbstractView.h
LoicCadour: Navigators() is enough, thx.

src/medCore/process/medAbstractProcess.h
LoicCadour: Let's keep it warm for next telco :)

rdebroiz:
I close this one, the navigator / interactors problem seems to be OK finally, and the process API will be totally redefined soon anyway...

PR #260, opened 2015-07-23
[Layer/View Settings] ergonomy, remove empty area

mathildemerle:
Hello everyone,

On the MUSIC side, i worked on "Layer settings" & "View settings" toolboxes ergonomy. These toolboxes were too much large, and we wanted to be able to hide everything.

So :
>1° Remove empty area in View settings and Layer settings
>2° Remove empty line at the end of the data list
>3° Link interactorToolbox with Layer settings: when we want to hide Layer settings, it hides also interactorToolbox.
>4° I don't know you, but the contour of the groupBox of interactorToolbox was a bit ... ugly ^^'. I removed it.
>5° Link menu position changed. In my humble opinion i think it's better to put it like that, the user will see that they are specific buttons. 

It's my first PR on your side, so, tell me if it's (not) good, if you (don't) like it, etc :)

>PS: :rabbit:


rdebroiz:
It indeed looks better IMO. Good for me.

before:

before

after:

after

mathildemerle:
Thanks @rdebroiz  :)

PR #263, opened 2015-09-09
Add, 'add', 'divide' and 'multiply' images process.

rdebroiz:


NicolasSchnitzler:
Hello, great job.
I think it could be simpler though:
-dtk provides a macro to generate the XXXPlugin class automatically (unless the inheritance prevents it to work in your case?)
-i don't think the .cpp is required, the tags that are defined there are (in my understanding), for the composer. However, the node class provides a .json file, that defines these tags itself.

rdebroiz:
> -dtk provides a macro to generate the XXXPlugin class automatically (unless the inheritance prevents it to work in your case?)

Could you show me a dummy example of how to use them ?

> -i don't think the .cpp is required, 

Not sure to understand which cpp you're talking about.

> the tags that are defined there are (in my understanding), for the composer. However, the node class provides a .json file, that defines these tags itself.

Yes I know but those tags could be used somewhere else in the app as well. Is therean easy way to get the information from the json files (I know dtk used it for the composer, but I'm lazy to look for it by myself, if you know, can you tell me ;)) ?

NicolasSchnitzler:
> Could you show me a dummy example of how to use them ?

I have tried something in that way here: https://github.com/NicolasSchnitzler/medInria-public/tree/composer, more specifically the following files 

src/medCore/process/medAbstractGaussianFilter.h,   

src-plugins/meditkGaussianFilterPlugin/meditkGaussianFilterPlugin.cpp src-plugins/meditkGaussianFilterPlugin/meditkGaussianFilterPlugin.json 
src-plugins/meditkGaussianFilterPlugin/meditkGaussianFilterPlugin.h 

which is just a wrapper over (a small part) of the existing itkFilters plugin, to change its interface, so as to be able to integrate it in the composer.

> Not sure to understand which cpp you're talking about.

I am referring to the .cpp in  src/medCore/process/arithmetic_operation
In my understanding, these should be interfaces, so have .h only

> Yes I know but those tags could be used somewhere else in the app as well

Ok, didn't see it that way

> Is therean easy way to get the information from the json files

Qt provides some facilities to handle that (QJsonObject, QJsonDocument), for example, look at https://github.com/d-tk/dtk/blob/master/src/dtkComposer/dtkComposerNodeMetaData.cpp, method setFromFile

rdebroiz:
> I am referring to the .cpp in src/medCore/process/arithmetic_operation
In my understanding, these should be interfaces, so have .h only

OK it's just beacuse there is the tags method which is not virtual pure, maybe we will add more things in it later. It's also better for compilation time, if we add a new tag, we just have to recompile this .cpp and not all the .cpp including the header.

rdebroiz:
> I am referring to the .cpp in src/medCore/process/arithmetic_operation
In my understanding, these should be interfaces, so have .h only

Michael make me realise that there will have nothing in that file if we retrieve the tag list from the JSON file, so we can indeed prune it. I will update the PR soon.

rdebroiz:
Ok so i add a struct to embed all the details of `medAbstractProcess`:
```c++
struct medProcessDetails
{
    QString name;
    QString version;
    QStringList dependences;
};
```

I also add a utility function to fill this struc from a JSON file in medProcessLayer.cpp: 
```c++
medProcessDetails readDetailsFromJson(QString const& filePath);
```

It can be used to load it from a ressource in a plugin example from `medItkSubtractImagesPlugin`:
```c++
class medItkSubtractImageProcessPrivate
{
public:
    medProcessDetails details;
};

medItkSubtractImageProcess::medItkSubtractImageProcess(QObject *parent): 
    medAbstractSubtractImageProcess(parent),
    d(new medItkSubtractImageProcessPrivate)
{
    d->details = medProcessLayer::readDetailsFromJson(":/medItkSubtractImageProcessPlugin.json");
}
```

rdebroiz:
I tried to use the macro provided by dtk to create the plugin, pluginFactory and pluginManager classes but I got an error when I load the plugin, and I can't figure out from where it comes:

`libmedItkSubtractImageProcess.so: undefined symbol: _ZN37medAbstractSubtractImageProcessPlugin16staticMetaObjectE`

where the staticMetaObject is supposed to be defined ? 
@NicolasSchnitzler does that ring  a bell with you?

rdebroiz:
OK I found the problem, you have to explicitly write the Q_OBJECT macro inside the file that use the dtk macros, otherwise it is not detected by the automoc tool.

PR #264, opened 2015-09-09
Uses cmake properties system to set the output directories of target …

rdebroiz:
…rather than redefine the global variables each time.


rdebroiz:
Integrated in https://github.com/medInria/medInria-public/pull/265

PR #267, opened 2015-09-11
Add, morphomath images process.

rdebroiz:
- Erode
- Dilate
- Closing
- Opening


rdebroiz:
I also put all the old style plugin in a legacy folder and start to construct a hierarchy tree like the one of the abstractions in the Core layer.

PR #270, opened 2015-09-18
MODULE CMake Keyword

rdebroiz:
Use the MODULE CMake Keyword in add_library([...]) to produce lib having to be dynamically loaded (plugins).

see: http://www.cmake.org/cmake/help/v3.0/command/add_library.html

### DIFF
--- a/src-plugins/process/arithmetic_operation/medItkAddImageProcess/CMakeLists.txt
+++ b/src-plugins/process/arithmetic_operation/medItkAddImageProcess/CMakeLists.txt
@@ -52,7 +52,7 @@ include_directories(${${TARGET_NAME}_INCLUDE_DIRS}
 ## #############################################################################
 
 
-add_library(${TARGET_NAME} SHARED
+add_library(${TARGET_NAME} MODULE
   ${${TARGET_NAME}_CFILES}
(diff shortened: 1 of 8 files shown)


papadop:
Just for curiosity, what is the difference with SHARED ??? The documentation is not very clear on the difference between SHARED and MODULE.... Otherwise, that seems a reasonable move given the documentation you pointed to. On the other hand, there is the fanous "If it works don't change it...", but I'm not totally endorsing this view.

papadop:
And they are probably many other places where we want to make a similar change.

rdebroiz:
> The documentation is not very clear on the difference between SHARED and MODULE....

As far as I know there is no difference on linux or windows plateforms: http://stackoverflow.com/questions/4845984/difference-between-modules-and-shared-libraries. But It gives one more hint about how the library is supposed to be used

> And they are probably many other places where we want to make a similar change.

Yep I know I did this only for the plugin using the new systeme provided by dtk. I told myself : if we have to do those things again let's do it the right way. There's a mechanism in CMake to tell that you want to build a plugin, so let's use it ;)

rdebroiz:
About your last comment, maybe you didn't noticed that this PR will be merged on the `dev` branch.

PR #272, opened 2015-09-18
Composer

NicolasSchnitzler:
**require up-to-date dtk1.0 version (09-17-2015 or later)**
First version of the composer

All the nodes correspoding to the plugin itkFilters have been ported, with their corresponding abstraction
Using dtk widgetFactory, (hence the above requirement), plugins are able to provide their own widgets to the composer. (In a way that allows the process to be used - for tests notably - without GUI).

the new itkRigidReistration plugin demonstrate how to write a new plugin from scatch (other plugins being wrappers around legacy plugins)

the meditkRegistrationDiffeomorphicDemonsProcess node demonstrate how to make use of an implementation specific widget (depending on a plugin). 

User interaction is not available yet.


rdebroiz:
there is `#pragma once` in some cpp files, for example in `/src/medComposer/medErodeFilterNode.cpp`

NicolasSchnitzler:
corrected now, 
thanks for the insight

NicolasSchnitzler:
outdated, closing

PR #276, opened 2015-10-07
Dev new superproject

rdebroiz:


rdebroiz:
This incorporates the superbuild directly in the medInria repository.

Main advantages:
- no more headache synchronisation between the tow repositories (medInria-public, and medInria-superproject each time a third part library is updated

- no more mess with the medInria update step (no update step at all for medInria, the code is already here). So it's always safe to launch a build of the super-build, even if there is uncommitted modifications in medInria.

It require cmake 3.2 if there it is not packaged for your OS yet, you can compile it from the source : `git://cmake.org/cmake.git` (there are release tag/branches). It is need for some awesome new features such as the `BUILD_ALWAYS` flag of the `ExternalProject` module which force to always update->configure->build any external project. So we can get rid of our custom step target to force them on Jenkins.

Things are organised slightly differently than before, I use the `EP_PREFIX` flag to set the source/build/install directories of all the third part libraries. they are now download and builded in the build directory (in `${CMAKE_BINARY_DIR}/superbuild/src`). 

medInria is build in `${CMAKE_BINARY_DIR}/medInria-build`. It can still be configure and builded by its own (it is still an "external project" actually). To dev just open the VS solution generated in `${CMAKE_BINARY_DIR}/medInria-build` or the CMakeLists.txt in `${CMAKE_SOURCE_DIR}/src`

Qt5 components are found by providing the path to the `Qt5Config.cmake` file for the Qt5_DIR variable. on my machine it's `.../Qt/5.4/gcc_64/lib/cmake/Qt5`

The packaging and the get revision project are broken, It will be fixed soon.

I tested it with clang/gcc and make/ninja.

The composer Area made by Nicolas is integrated (accessible from the homepage area), there are nodes for the process present in the medCore layer.

PR #277, opened 2015-10-13
Dev mac

ocommowi:
Corrects compilation errors, warnings and link errors on OSX

Check needed that it still compiles well on linux


src/app/medInria/QSingleApplication/qtlocalpeer.cpp
```diff
@@ -93,7 +93,7 @@ QtLocalPeer::QtLocalPeer(QObject* parent, const QString &appId)
         socketName += QLatin1Char('-') + QString::number(sessionId, 16);
     }
 #else
-    socketName += QLatin1Char('-') + QString::number(QtLP_Private::getuid(), 16);
+    socketName += QLatin1Char('-') + QString::number(getuid(), 16);
 #endif
```
rdebroiz: > error: use of undeclared identifier 'getuid'; **did you mean 'QtLP_Private::getuid'?**
    socketName += QLatin1Char('-') + QString::number(getuid(), 16);
                                                     ^~~~~~
                                                     QtLP_Private::getuid
/usr/include/unistd.h:675:16: note: 'QtLP_Private::getuid' declared here
extern __uid_t getuid (void) __THROW;

**LOL**

rdebroiz: i add `#include <unistd.h>`

rdebroiz:
There is no more QtSingleApplication provided by Qt since it use a lot of classes deprecated in Qt5, however there is mirror here: https://github.com/qtproject/qt-solutions where the code seems to be maintained and works with Qt5. But It's coming with lot of useless things for us.

Otherwise another (better) alternative could be this project which another implementation of a Qt single application: https://github.com/itay-grudev/SingleApplication

I propose to keep it like that if it works on mac with my last modification, and keep `itay-grudev/SingleApplication` in mind for later.

ocommowi:
The app compiles and runs on mac, merging

PR #280, opened 2015-10-23
process without GUI.

rdebroiz:
Lot of files have been moved, all old medInria libraries are located in a legacy folder.
old medXXXparametre and medXXXJob are suffixed wit a capital 'L' to avoid conflict with new ones.

Start implementing new parameter that don't directly exposed GUI component.

Processes don't directly exposed GUI component they will be via medXXXPresenter classes.

Processes do not inheriths directly from `QObject` and `QRunable` anymore but from `medAbstractJob`. Classe that derived from `medAbstractJob`are registered to the `medJobManager` singleton. So all running jobs can be accessed from the application, in order to present all of them in the same view, or cancel all of them at the end of the application for example. 

All arithmetic operation processes derived types are shipped in only one plugin, `medItkArithmeticOperationProcessPlugin`

All morphomath operation processes derived types are shipped in only one plugin, `medItkMorphomathOperationProcessPlugin`.

Tow new classes have been introduced: `medArithmeticOperationProcessNode<T>` and `medMorphomathOperationProcessNode<T>`, for factorisation purpose.


papadop:
Is this the branch to look at... I suspect not.
But if yes, can you give some clues about where to look at.

 Because there are 664 file changes with lots of non-essential stuff (such as commented lines -- by the way prefer #if 0 # endif to comments, less changes, more readable, or completely remove the lines ---), so that it is not super easy to see the crux of the proposal...

I do not find any instance of medItkArithmeticOperationProcessPlugin or medItkMorphomathOperationProcessPlugin in the file tab... so I'm a little bit puzzled. I found a medTestWorkspace, which refers to filtering stuff, but that's it.

ocommowi:
Actually RP has worked a lot by doing pull requests to the dev branch. So this is only part of the work. You should look at the dev branch on medinria/medinria-public. That being said it does contain an awful lot of changes including file moves, refactoring of the superproject (you don't need medinria-superproject with the dev branch, etc. But the abstract part of process refactoring is mainly in the folder layers/medCore (job and process particularly). Then, you can find plugins using that in the plugins folder, I remind that mainly the morpho math plugin is the one that is implementing both a presenter and a process.

papadop:
So can we close this one as it is merged...

ocommowi:
Well, in fact it is merged so yes it is closed :)

PR #286, opened 2015-12-02
command line interface

NicolasSchnitzler:
Add the possibility to use medInria from command line without loading the GUI.

Only the new style processes (abstraction based) are supported. 
For each abstraction a medXXXCommandLineExecutor is defined, allowing to handle the abstraction specific parameters. 
The abstraction to use is defined by the -a option, the plugin used can be forced using -p. 

Example: 

./medInria --no-gui -a subtract -r img1.nii -l img2.nii -o res.nii


NicolasSchnitzler:
outdated. 
Furthermore, a template based implementation could most likely avoid duplicated code

PR #291, opened 2016-01-26
Dev3

mikejbuckingham:
Don't merge, obviously. Just, you know, observe.


ocommowi:
Hey there, I tried compiling your branch just as a starter. I get a lot of errors like this one : 

/Users/ocommowi/MedINRIA-NT-Code/medinria/src/app/medInria/medGenericWorkspace.cpp:109:48: error: cannot initialize a parameter of type 'medAbstractAddFilterProcess *' with an lvalue of type 'medAbstractProcess *'
            return addFactoryPresenter->create(myProcess);

I guess this is related to what you mentioned last time and to my email. I see two solutions to that. Either we explicitly cast myProcess to the right abstraction before calling create; or we modify the process presenter factory so that the create method takes a medAbstractProcess * (I think it is doable, the cast to the abstraction would be in the create method instead of the generic workspace). Tell me what you think

ocommowi:
I also rebased your branch on the original dev branch. Do you want me to force push it instead of yours (so that everything stays up-to-date) ?

mikejbuckingham:
Sure - should have done that myself. 

----- Mail original -----

mikejbuckingham:
We can modify the presenter, although we are increasingly reliant on QVariant which makes me uncomfortable as it's hidden in the code. 

An alternative is just to, you know, *not* use the hack I did and do things long-hand, it's just a bit frustrating to type out. 

----- Mail original -----

ocommowi:
So I force pushed the rebase and the little modification in the presenter factory create methods. I don't see any qvariant coming up right now but we can discuss that this afternoon.

I still have errors there but that's because you register the arithmetic operations one factory too high (any reason for that apart from simplifying the generic workspace?) : 

/Users/ocommowi/MedINRIA-NT-Code/medinria/src/plugins/process/arithmetic_operation/medItkArithmeticOperationProcessPlugin.cpp:28:72: error: cannot initialize a parameter of type 'creator' (aka 'medAbstractArithmeticOperationProcess *(*)()') with an lvalue of type 'medAbstractAddImageProcess *()': different return type ('medAbstractArithmeticOperationProcess *' vs 'medAbstractAddImageProcess *')
                                                                     , medItkAddImageProcessCreator);

Issue #292, opened 2016-02-18
Tests (update please as we go)


mikejbuckingham:
So as requested, can we get people to test the following:

medInria compiles on Windows, Linux and MacOS
medInria correctly loads all plugins

All medInria filters work as expected:
    Add
    Divide
    Gaussian
    Invert
    Multiply
    Median
    Normalize
    Subtract
    Windowing

Visualiation works as expected (2d, 3d) on all OSes and both Intel and nVidia chipsets
Visualisation performance has not significantly degraded from VTK5 (medInria 2.2.1)

Composer provides full functionality
Presenter vs Port for parameters correctly values Port over Presenter
All nodes have the correct ports and presenters
Compositions work as expected, end-to-end

Please also play around with the app and report all crashes, weird behaviours, etc.

ocommowi:
Hey,  I had been playing around with a previous version and noted quite a few bugs in issue #261 . Could you merge those two issues (and may be tick what you have finished, I believe there are many) ? Thanks and great job by the way !

ocommowi:
First list of bugs from my side (moved from #261)

  - [ ] Paint segmentation does not work anymore with VTK7 compatible views (image is created without a name and painting is not there)
  - [ ] saving a loaded file to database makes its thumbnail disappear (written in the right folder but not found later on)
  - [ ] closing the last view of a tab leaves it in an awkward situation: not going back to base mode of the tab, just leaving it empty and not closable
  - [ ] the composer does not appear in bottom left menu and neither in all places where areas and workspaces are listed
  - [ ] Composer stylesheet to update
  - [ ] Ctrl + Space menu missing icon for generic workspace, composer missing as well
  - [ ] Going to the finder in browser on Mac makes the app crash
  - [ ] home page : generic does not appear in any category, clinical is empty
  - [ ] dropping a tensor image (or any glyph based image) in an empty view generates a lot of warnings in the console
  - [ ] stylesheet of combo box is wrong (eg selecting a plugin), probably some others as well
  - [ ] Diffusion tractography fails when a tensor is negative (TTK related, I'll handle it)
  - [ ] Generic, filtering and diffusion workspaces probably do something wrong when loading plugin processes or handling them: when quitting the app after using one plugin in those workspaces, there is a crash
  - [ ] Regression in diffusion workspace: in diffusion estimation, we cannot anymore do a brain masking plus model estimation. This would require an abstraction on its own, not sure yet how to handle this
[...]

PR #303, opened 2017-09-29
View bug correction with good base

Florent2305:
BugCorrection #292
Two bugs are fix:

Some view interaction glitches: for example drop the 4d image of the heart, then the heart mesh. If you interact with the view, it will change the window level of the first layer (image) even though the mesh is selected on the right. In addition if that view is linked to others (eg do a 4 view split after dropping the two data), then this window leveling is not synchronized with the others
Drop the 4d heart image, switch to 3d mode, click on play in the navigator -> the image doesn't move. It does work in 2d but not 3d anymore...
Correction consists of:
-replace SetInputData by SetInputConnection;
-Create an individual windowLevel filter for x,y,z actors.


ocommowi:
Perfect, working -> merge on the way

PR #316, opened 2018-01-26
About plugin

Florent2305:
Correction of We've "lost the about plugins buttons. We should put them back" by mikejbuckingham


src/layers/legacy/medCoreLegacy/gui/toolboxes/medRegistrationSelectorToolBox.cpp
```diff
@@ -202,6 +205,9 @@ void medRegistrationSelectorToolBox::changeCurrentToolBox(int index)
     connect (toolbox, SIGNAL (success()),this,SLOT(enableSelectorToolBox()));
     connect (toolbox, SIGNAL (failure()),this,SLOT(enableSelectorToolBox()));
 
+    this->setAboutPluginVisibility(true);
+    connect(this->header()->aboutButton(), SIGNAL(triggered()), toolbox, SLOT(onAboutButtonClicked()));
```
ocommowi: Manque un disconnect quelque part. Car la en changeant de plugins, c'est l'avalanche de popups

Issue #327, opened 2018-02-15
Crash when linking views


ocommowi:
Here comes a new crashing bug 

- load an image and drop it in two views next to each other
- select to link them (ctrl+click)
- link the views on the right side
- switch to 3d and zoom (right click on view)

And crash ! Apparently because of an infinite loop in parameters

PR #329, opened 2018-02-20
Fix "Crash when linking views" #327

Florent2305:
Fix "Crash when linking views" #327


ocommowi:
Perfect, je ferme le bug

Issue #330, opened 2018-02-21
Release stuff issue


ocommowi:
A bunch of final things that would be nice to do before release:

- [x] Interactor option for itkDataImage to switch between interpolation modes (nearest / linear)
- [x] Drag and drop files from OS (windows at least)
- [x] Resizing splitter when closing views (size problem if not handled)

Optional:
- [x] Tensor estimation plugin (multi-shell)
- [ ] Image arithmetic plugin templating #320

PR #331, opened 2018-02-21
1st step for interpolation button

Florent2305:
1st step for "Interactor option for itkDataImage to switch between interpolation modes (nearest / linear)" into #330


ocommowi:
Works for me, merging

PR #333, opened 2018-02-27
Add "drag and drop" mechanism for externals files

Florent2305:


ocommowi:
Alirght, it works well ! And no modifications of medDataManager as we were afraid of. Good to merge

PR #350, opened 2019-05-06
Mesh LUT Correction

Florent2305:


ocommowi:
Corrects bug when showing scalar maps on surfaces very well

PR #352, opened 2019-06-03
[music3] DCMTK Dicom import debug

mathildemerle:
(From https://github.com/Inria-Asclepios/medInria-public/pull/571)

This PR fixes the DICOM import error from DCMTK 3.6.2 in medInria.

:m:


Florent2305:
It's Ok

PR #354, opened 2019-06-03
[music3] default thumbnail right path 

mathildemerle:
(From https://github.com/Inria-Asclepios/medInria-public/pull/22)

This PR can/will be rebased with https://github.com/medInria/medInria-public/pull/352

The path to the default thumbnail image was wrong.

:m:


Florent2305:
Good

PR #371, opened 2019-07-05
[music3] DB cleanup and debug, import mha and dicom

mathildemerle:
This is a PR about some debugs/enhancement that we have done over the years about the data import/database. It is necessary to import mha in the application for instance.

 - Import mha and DICOM debuged
 - Remove unnecessary logs
 - Metadatakey -> Description is used to add the name and version of the app used to create the data
 - MRI -> add metadata FlipAngle, EchoTime, RepetitionTime
 - addmetadata -> use setmetadata
 - Handling of errors
 - Handle extensions and priorities in writers
 - The DB path was wrong (QT5: missing "/data/appname")
 - Add the filter toolbox on top of the database column in workspaces
 - Right click on a series in the DB -> Metadata -> shows the list of metadata
 - Debug the thumbnail size which is crazy sometimes in the DB column
 - In the DB column, F2 (or other key according to your OS) allows to rename a patient/study/series
 - Start the debug of the thumbnail disappearance in DB when you save a data. The metadata is not saved in the database, even if everything look fine. 

Do not hesitate to look at the changes, in case i wrongly mixed music and medInria code :)

Commit: https://github.com/medInria/medInria-public/commit/3bafc5afd2bd34c0d010ca55455cb56f8d6a7fff

:m:


mathildemerle:
This PR has been rebased

src/layers/legacy/medCoreLegacy/database/medDatabaseProxyModel.cpp
```diff
@@ -62,41 +62,47 @@ void medDatabaseProxyModel::clearAllFilters()
...
 {
-
     // get the current model index
     QModelIndex current(sourceModel()->index(source_row, 0, source_parent));
     // get the data we want to check
     QModelIndex index = sourceModel()->index(source_row, currentKey, source_parent);
 
     // show all children if parent is valid
-    if (current.parent().isValid() && !isCheckingChild && !isCheckingParent)
+    if (current.parent().isValid() && (isCheckingChildren == 0))
```
Florent2305: !isCheckingChild is better than (isCheckingChildren == 0)

src/layers/legacy/medCoreLegacy/database/medDatabaseProxyModel.cpp
```diff
@@ -62,41 +62,47 @@ void medDatabaseProxyModel::clearAllFilters()
...
+            isCheckingParents--;
             return true;
+        }
+
+        isCheckingParents--;
     }
 
     // show the parent if one child is valid
-    if(sourceModel()->hasChildren(current) && !isCheckingChild && !isCheckingParent )
+    if(sourceModel()->hasChildren(current) && (isCheckingParents == 0))
```
Florent2305: !isCheckingParents is better than (isCheckingParents == 0)

src/layers/legacy/medCoreLegacy/database/medDatabaseProxyModel.cpp
mathildemerle: isCheckingChildren is not a boolean here. The changes are a bit old, but have been made by @ocommowi  and @fcollot in this commit https://github.com/Inria-Asclepios/medInria-public/commit/676758363b5aec601098c24ae09cce24864a2b41#diff-8a3c89be0d75ff3017f2e8437c78aebb

mathildemerle: (if you do not have access to this commit, its name is "Enable multiple depth levels instead of two in search" and change only that part)

Florent2305: Ok je savais pas
Bizarre un is qui est pas bool...

src/layers/legacy/medCoreLegacy/database/medDatabaseProxyModel.cpp
Florent2305: Identique au commentaire précédent.

src/layers/legacy/medCoreLegacy/database/medDatabaseProxyModel.cpp
fcollot: (I merged that PR, but I did not work on it)

fcollot:
The _msc_ prefix is specific to MUSIC and should not be present un the medInria code. Any code transfered from MUSIC to medInria should have its prefix changed to _med_.

mathildemerle:
Yes, there is an issue already about that. Other files (from medUtilities, or mscAlgorithmPaint) are concerned. I'm going to do a PR with these changes after. https://github.com/medInria/medInria-public/issues/394

fcollot:
There are too many unrelated issues addressed in this PR to properly review it.
Anyways, if this is all copied from our fork without modification of the code then there is no need to review what we've already reviewed in the past, so if there isn't anything new here then I'm ok with it.

mathildemerle:
Indeed @fcollot there are really too many issues related to this PR. The thing is: it as been a nightmare to extract years and years of debug from our side about the database. We have been changing the code everywhere. I've stopped when i had a working state of code in fact. Thx for the review

PR #384, opened 2019-07-29
External project plugins

fcollot:
In MUSIC 2 we have various private plugin repositories, and setting up a working MUSIC development environment is impractical. Each private set of plugins must be cloned separately, and each has its own superproject which needs to be configured with the path to the medInria superproject, and inversely the medInria superproject needs to be configured with the path of the private plugins to load them.
It is much simpler to remove the plugin superprojects and set them up as external projects, so that medInria handles their configuration automatically. I did this successfully with the main MUSIC plugins repo, and it requires the addition of a cmake variable `EXTERNAL_PROJECT_PLUGINS_LEGACY_DIRS` that allows external project configuration files to add the path of their plugins.

_(The inclusion of MUSIC as an external project will be done on our fork of medInria)_


Florent2305:
It's smell good.

PR #400, opened 2019-09-19
[music3] generateThumbnail - remove pop-up at import in DB

mathildemerle:
From this PR https://github.com/Inria-Asclepios/medInria-public/pull/569

The original problem in some Linux drivers seems to be removed since Ubuntu 18. It was there on Ubuntu 16 on X.org drivers, not nvidia. On Ubuntu, these driver names have changed, and always start with "Intel". 

The non-offscreenCapable algo in this medAbstractData::generateThumbnail creates a pop-up when you save a data in a toolbox for instance. So, using the offscreenCapable removes it.

:m:


mathildemerle:
PR rebased

PR #403, opened 2019-09-19
[music3] four-split and maximized actions menu cleaned

mathildemerle:
From https://github.com/Inria-Asclepios/medInria-public/pull/259

 * Four-split button becomes a menu action in "Tools"
 * Maximized menu action used "toggle" to display if the view is maximized or not
 * clean the interactions with/without data

:m:


mathildemerle:
This PR solves the crash opening four-split view with several layers

PR #409, opened 2019-09-25
Improve binaries outputs directories

Florent2305:


src/CMakeLists.txt
```diff
@@ -98,7 +98,7 @@ include(set_exe_install_rules)
 include(set_lib_install_rules)
 include(set_lib_properties_variables)
 include(set_plugin_install_rules)
-include(set_plugin_install_rules_legacy)
+#include(set_plugin_install_rules_legacy)
```
mathildemerle: this line could be removed?

src/cmake/module/set_plugin_install_rules.cmake
```diff
@@ -11,31 +11,85 @@
...
+################################################################################
+#
+# Usage: set_plugin_install_rules_legacy(target)
+# set rules for the plugin legacy designed by the target
+#
+################################################################################
+get_property(GENERATOR_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
+
+if(${GENERATOR_MULTI_CONFIG})
+  set_target_properties( ${target} PROPERTIES RUNTIME_OUTPUT_DIRECTORY_DEBUG          ${CMAKE_BINARY_DIR}/${platformType}Debug/bin/plugins_legacy)
```
mathildemerle: So, you want to keep the plugins in the binary directory? Does it removes the original "plugins" directory at the package root?

Florent2305: No decision has been made yet.
We were supposed to talk about it during Monday's videoconference but we did not have time to tackle this point, unfortunately.
I propose to keep this behavior, until we decide for the v3.1 then we fix it all in a dedicate PR.

mathildemerle: Ok then, it's not a critic problem, it's just a bit weird if you search plugins in the linux package since there are 3 plugins directories. But indeed, it's not an emergency. Perhaps i can create an issue about it to remember the problem?

Florent2305: Ok I create the issue #410

src/CMakeLists.txt
Florent2305: it's done

PR #428, opened 2019-10-10
[music3] Meshing Workspace + 4 toolboxes

mathildemerle:
This Pull Request adds:
 * the _Meshing_ workspace
 * the _Create Mesh from Mask_ toolbox to convert a mask to a mesh
 * the _Mesh Manipulation_ toolbox to manually scale/rotate a mesh
 * the _Mesh Mapping_ toolbox to map a data to a mesh
 * the _Remeshing_ toolbox to refine/decimate/smooth/choose the triangles number of a mesh

Small comments: 
 * medComboBox.h handle the way the selector combobox in workspaces behaves. It's from https://github.com/LoicCadour/medInria-public/pull/82 but the wheel event is removed. Indeed, the user can loose his/her work if the wheel event in a toolbox is caught by the selector.
 * the change from dtkDebug to qDebug in medDatabaseView.cpp  is a debug: the soft freezed sometimes saving data in the db.

:m:

### DIFF
--- a/src/layers/legacy/medCoreLegacy/database/medDataManager.cpp
+++ b/src/layers/legacy/medCoreLegacy/database/medDataManager.cpp
@@ -372,17 +372,20 @@ void medDataManager::garbageCollect()
     }
 }
 
-
 QUuid medDataManager::makePersistent(medAbstractData* data)
 {
     if (!data)
+    {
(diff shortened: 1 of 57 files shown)


mathildemerle:
PR rebased :)

PR #435, opened 2019-10-15
[CMakeLists] add required components

mathildemerle:
Add libraries components for ITK and VTK in plugins.

:m:


Florent2305:
It's good for me ! ;)
Is it not too difficult to do?

mathildemerle:
It was ok. I had to open VTK and ITK libraries in QtCreator. Each time i run the compilation, i had a file error with VTK and ITK. I just needed to open the corresponding file, and the module was written in its header. So it was just a bit repetitive :)

PR #444, opened 2019-10-17
[Legacy Workspaces] clean the use, add logo

mathildemerle:
Seeing that the new "selector" workspaces did not had logo (CTRL-SPACE on the application), i looked at the creation of legacy workspaces. 

They are everywhere: some in `app/medInria`, some in `app/medInria/[...]/workspaces`, some in the legacy plugins. Some of them were called in `medApplication::initialize()` to be registered, some others were registered by their plugins.

So, i added the missing legacy workspace logos (Meshing, Reformat, Filtering Legacy), and moved FilteringL, Registration and Segmentation to the legacy plugins. They are all registered by their plugins, and it allows to add workspaces in external projects. If it's a problem, i can put them back, do not hesitate. The Visualization workspace still need to be in the "core" because it's called in `medMainWindow::open`. This is for opening through command-line as far as i understand. This workspace is a bit special, and is not a selector workspace.

:m:


mathildemerle:
@Florent2305 according to your suggestions, i've changed the workspace name of "Filtering Legacy" to "Filtering", and "Filtering" to "Filtering dtk2". And it will be quicker to update music2 plugins to medInria3.1.

PR #449, opened 2019-10-18
fix bug "Filtering process creation without process #408"

### LINKED ISSUE #408: Filtering process creation without process
Into "new" filtering workspace.
If at least one filter exist and you change "Process Type" then "Process" combobox is set to "None selected" but you can create a process

Florent2305:
fix #408

### DIFF
--- a/src/app/medInria/areas/workspaces/filtering/medFilteringWorkspace.cpp
+++ b/src/app/medInria/areas/workspaces/filtering/medFilteringWorkspace.cpp
@@ -354,8 +354,9 @@ class medFilteringWorkspacePrivate
 public:
     int iProcessSelection;
 
-    QComboBox *processTypeComboBox;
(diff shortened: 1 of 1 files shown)


src/app/medInria/areas/workspaces/filtering/medFilteringWorkspace.cpp
```diff
@@ -449,6 +450,9 @@ void medFilteringWorkspace::setProcessType(int index)
     d->morphoMathsVector.clear();
     d->arithmeticsVector.clear();
 
+    if (index == 0)
+    return;
```
mathildemerle: there is an indentation problem here

Florent2305: it's  done

PR #459, opened 2019-10-23
[PolygonROI] crash delete roi and switch to other tlbx

mathildemerle:
Fix the crash:
 * Drop a data in a view, open PolygonROI
 * Draw a ROI, delete it (with the "D" button in medRoiManagementToolbox)
 * Switch to an other toolbox (Paint for instance) -> crash

:m:


mathildemerle:
@Florent2305 done :)

PR #467, opened 2019-10-29
[N4Bias] Filtering tlbx medN4BiasCorrection added

mathildemerle:
This PR adds the N4 Bias Correction toolbox in the Filtering (legacy) workspace.

:m:

### DIFF
--- a/src/plugins/legacy/medN4BiasCorrection/CMakeLists.txt
+++ b/src/plugins/legacy/medN4BiasCorrection/CMakeLists.txt
@@ -1,0 +1,83 @@
+################################################################################
+#
+# medInria
+#
+# Copyright (c) INRIA 2013 - 2019. All rights reserved.
+# See LICENSE.txt for details.
+# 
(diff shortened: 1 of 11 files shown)


src/plugins/legacy/medN4BiasCorrection/CMakeLists.txt
```diff
@@ -0,0 +1,83 @@
...
+  ${${TARGET_NAME}_CFILES}
+  ${${TARGET_NAME}_QRC}
+  )
+
+## #############################################################################
+## Link
+## #############################################################################
+
+target_link_libraries(${TARGET_NAME}
+  ${QT_LIBRARIES}
```
Florent2305: Is it enough ? no ITK libs ...
Is it done by line 36?

mathildemerle: The compilation does not ask anything here. Do you think you're going to have a problem here on Windows?

PR #475, opened 2019-10-31
[VariationalSeg] interactors cleaning

mathildemerle:
This PR aims to clean/solve remaining problems seen in Variational Segmentation from the interactors system.

 * No crash closing the last layer of the view in the toolbox
 * Slicing Parameter standardize in interactors

And something weird in varSegToolBox.cpp:
if you do not save the output data in the db (with `VarSegToolBox::addBinaryImage`) before adding the output in the view, it crashes. During this crash, `medVtkViewItkDataImageInteractor::setInputData` is called, but does not succeed to do `SetViewInput (data, d->view->layer(data)`.
I have got a `"Unable to add data:  \"itkDataImageUChar3\"  to view  \"medVtkViewItkDataImageInteractor\""` warning.
Thus, the data is not really created, including parameters which create the crash. 
In this PR, since we want to do an alpha release soon, i choose to save the data (non-permanently) in the database before adding the output on the view. However, it could be an underlying problem. The `SetViewInput` function is a new function which was different from the [previous one](https://github.com/Inria-Asclepios/medInria-public/blob/2.4.0/src-plugins/itkDataImage/interactors/medVtkViewItkDataImageInteractor.cpp#L158) so i don't know how to solve this. I checked and this behavior was not bugged on our version.
I think this is the only toolbox for now trying to add data like that, that's why it has not been seen before.

:m:

### DIFF
--- a/src/plugins/legacy/variationalSegmentation/varSegToolBox.cpp
+++ b/src/plugins/legacy/variationalSegmentation/varSegToolBox.cpp
@@ -219,6 +219,8 @@ void VarSegToolBox::applyMaskToImage()
 {
     if (d->currentView)
     {
+        addBinaryImage();
+
         this->setToolBoxOnWaitStatus();
 
         d->process = qobject_cast<medAbstractProcessLegacy*>(dtkAbstractProcessFactory::instance()->create("medMaskApplication"));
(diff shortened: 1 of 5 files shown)


src/plugins/legacy/variationalSegmentation/varSegToolBox.cpp
```diff
@@ -481,7 +483,7 @@ void VarSegToolBox::endSegmentation()
         enableButtons(false);
 
         medAbstractView *view = this->getWorkspace()->tabbedViewContainers()->getFirstSelectedContainerView();
-        if (view)
+        if (view && (view->count()>0))
```
Florent2305: view is not enough ?

PR #479, opened 2019-11-05
[FIX] - [OSX] - default db path

juliencastelneau:
On OSX, Replace the default DB path below  : 
- `/Users/account_name/Library/Application SupportmedInria`

with
- `/Users/account_name/Library/Application Support/inria/medInria/`

**CAUTION**, 
If you have already launch medInria on your laptop, 
the wrong db path is defined in your plist 
cf. key `database.actual_database_location` in file `/Users/account_name/Library/Preferences/com.fr.medInria.plist`
and the fix does not work.

**First, You have to delete the key from your plist**
1. Open a terminal
2. `defaults delete com.fr.medInria database.actual_database_location`

### DIFF
--- a/src/layers/legacy/medCoreLegacy/database/medStorage.cpp
+++ b/src/layers/legacy/medCoreLegacy/database/medStorage.cpp
@@ -60,8 +60,8 @@ QString medStorage::dataLocation(void)
         {
 #ifdef Q_OS_MAC
             vDbLoc = QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation)
-                .remove(QCoreApplication::applicationName())
-                .append(QCoreApplication::applicationName());
+                    + "/" + QCoreApplication::organizationName()
+                    + "/" + QCoreApplication::applicationName();
 #else
             vDbLoc = QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation)
                     + "/data/" + QCoreApplication::organizationName()


ocommowi:
This one surprises me a lot. I have never had that problem before. I have tried removing everything medinria in the library folder (inria folder in application support and the plist file in preferences). When I rerun medinria with the current version (without your fix) the folder is the right one : /Users/account_name/Library/Application Support/inria/medInria/

Not sure where your problem came from but I would be curious if you could do the same test with the nightly build (med.inria.fr/downloads/nightly/latest). That sounds very weird to me that two macs with the same software have different behaviors

mathildemerle:
(if i can help, there were problems with `QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation)` in Qt5 which did no include "/", whereas the previous version of that method in Qt4 included it. So, as far as i remember, i needed also to change these kind of lines in the `else` of this function)

juliencastelneau:
The actual code (before my PR) is 
            vDbLoc = QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation).remove(QCoreApplication::applicationName()).append(QCoreApplication::applicationName());

As you can see, the organizationName (inria) is never used.
So, it seems impossible to write at /Users/account_name/Library/Application Support/inria/medInria/
Even if the « /«  is not the main issue.

ocommowi:
I have read the code before and after. I have not found what the QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation) is. 

But : it works on my two Macs without your change. It does use the organization name and application name. Soto me the question is : what does the previous code output by default and why is it different for me and for you ??? Thus my question : does medinria nightly on the website work on your mac after cleaning your library folders ?

mathildemerle:
The problem is explained on https://doc-snapshots.qt.io/qt5-5.11/qdesktopservices-obsolete.html for instance:

> Note: when porting QDesktopServices::DataLocation to QStandardPaths::DataLocation, a different path will be returned.
> 
> QDesktopServices::DataLocation was GenericDataLocation + "/data/organization/application", while QStandardPaths::DataLocation is GenericDataLocation + "/organization/application".

Before, medInria used `QDesktopServices::DataLocation`.

juliencastelneau:
I tried the nightly. 
It works only if your plist has already the good value for the key database.actual_database_location

Have you tried to clean your plist as I explained in the PR.

When you launch medInria, if there is a value in the plist, the code with QDesktopService is not executed anymore.

But, if it is your first launch and the plist has no value for the key database.actual_database_location, the code is executed.

In all case, the code below has no sense
            vDbLoc = QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation)
                .remove(QCoreApplication::applicationName())
                .append(QCoreApplication::applicationName());

Why do we removed the applicationName, then re-append it ?

ocommowi:
I made a small executable test and it confirms the behavior you had. So let's go with this one, sorry for the mess. I am sure I have removed all preferences files and it was working.

So ok for me

Issue #486, opened 2019-11-12
Memory issues when getting rid of images


ocommowi:
medInria seems to never release memory for image data loaded. For example, just do the following:

- temporary import an image from the browser area
- remove it from the database right after

The memory used by medinria will have grown by a significant amount, meaning the image (and may be even the viewers used to get the database thumbnails) are not destroyed.

If you do this with large images, medinria is quite fast using the whole memory available

ocommowi:
I think I found a good part of that one while looking at the itkDataImage interactor (medVtkViewItkDataImageInteractor.cpp). There is a m_poConv that never gets deleted and it contains : the converter between itk and vtk so I think nothing gets deleted.

I see two solutions : adding the delete where they are needed (SetViewInput and ~medVtkViewItkDataImageInteractor) or smart pointing this m_poConv. 

Not sure this is all but that surely does not help. It could be in other interactors as well so we need to check all image data plugins

Florent2305:
I made some changes about m_poConv, but it not solves the problem.
It done just a small improvement. 
When an image is load,  vtk (asked by medInria) creates 3 big buffer. After image is removed 2 big buffer stay in place.
![vs_memory](https://user-images.githubusercontent.com/7893917/69288369-a7e84200-0bf9-11ea-8cac-67ea4ca6cfd8.png)

Florent2305:
valgrind don't provide more informations for the moment

Florent2305:
A valgrind stack trace about the memories leaks
[valgrind_summarize.txt](https://github.com/medInria/medInria-public/files/3873891/valgrind_summarize.txt)

PR #496, opened 2019-11-21
itkvtk converter deletion 

Florent2305:
first step to correct   " Memory issues when getting rid of images #486 "

m_poConv correction.


mathildemerle:
Tell me if/when i need to re-review this PR :)

Florent2305:
It's a too small change to bother you.

Issue #493, opened 2019-11-19
[Warnings] linux compilation warnings


mathildemerle:
I've got a lot of these:

```
[ 96%] Building CXX object app/medInria/CMakeFiles/medInria.dir/medApplication.cpp.o
In file included from /home/mathildemerle/Dev/MUSIC3/medInria-public_ExtProjs/src/dtk/src/dtkCore/dtkCorePlugin.h:23:0,
                 from /home/mathildemerle/Dev/MUSIC3/medInria-public_ExtProjs/src/dtk/src/dtkCore/dtkCore:8,
                 from /home/mathildemerle/Dev/MUSIC3/medInria-public/src/layers/medCore/process/medAbstractProcess.h:18,
                 from /home/mathildemerle/Dev/MUSIC3/medInria-public/src/layers/medCore/medCore.h:16,
                 from /home/mathildemerle/Dev/MUSIC3/medInria-public/src/app/medInria/medApplication.cpp:17:
/home/mathildemerle/Dev/MUSIC3/medInria-public_ExtProjs/src/dtk/src/dtkCore/dtkCorePluginManager.h:62:0: warning: ignoring #pragma mark  [-Wunknown-pragmas]
 #pragma mark -
 
/home/mathildemerle/Dev/MUSIC3/medInria-public_ExtProjs/src/dtk/src/dtkCore/dtkCorePluginManager.h:63:0: warning: ignoring #pragma mark Manager [-Wunknown-pragmas]
 #pragma mark Manager Management
 
/home/mathildemerle/Dev/MUSIC3/medInria-public_ExtProjs/src/dtk/src/dtkCore/dtkCorePluginManager.h:76:0: warning: ignoring #pragma mark  [-Wunknown-pragmas]
 #pragma mark -
 
/home/mathildemerle/Dev/MUSIC3/medInria-public_ExtProjs/src/dtk/src/dtkCore/dtkCorePluginManager.h:77:0: warning: ignoring #pragma mark Plugin [-Wunknown-pragmas]
 #pragma mark Plugin Management
 
/home/mathildemerle/Dev/MUSIC3/medInria-public_ExtProjs/src/dtk/src/dtkCore/dtkCorePluginManager.h:82:0: warning: ignoring #pragma mark  [-Wunknown-pragmas]
 #pragma mark -
 
[...]

PR #505, opened 2019-11-25
[Warnings] remove a lot of medInria warnings

mathildemerle:
This PR solves almost every warnings listed in https://github.com/medInria/medInria-public/issues/493.

The only one missing from this PR is `warning: "vtkRenderingCore_AUTOINIT" redefined`. These warnings could be solved if we remove the components list in VTK `find_package` in plugins using VTK. This need to be investigated. 

:m:


superbuild/projects_modules/medInria.cmake
```diff
@@ -90,19 +90,6 @@ set(cmake_args
   -DBUILD_ALL_PLUGINS=OFF
   -DBUILD_COMPOSITEDATASET_PLUGIN=OFF
   -DBUILD_EXAMPLE_PLUGINS=OFF
-  -DBUILD_LEGACY/ITKDATADIFFUSIONGRADIENTLIST_PLUGIN=ON
```
Florent2305: Why do you remove these lignes?

src/plugins/legacy/itkDataSHImage/management/vtkSphericalHarmonicSource.cxx
```diff
@@ -253,11 +254,14 @@ void vtkSphericalHarmonicSource::PrintSelf(ostream& os,vtkIndent indent) {
        << NumberOfSphericalHarmonics << std::endl;
     os << indent << "SH Basis: " << NumberOfSphericalHarmonics << "x"
        << sphereT->GetOutput()->GetNumberOfPoints() << std::endl;
+
     os << indent << "SH Coefficients:" << std::endl << '[';
     for (int i=0;i<NumberOfSphericalHarmonics;++i)
+    {
         os << indent << SphericalHarmonics[i] << " ";
-        os << indent << ']' << std::endl;
```
Florent2305: Olivier are you agree with this change?

superbuild/projects_modules/medInria.cmake
```diff
@@ -54,7 +54,7 @@ if (NOT USE_SYSTEM_${ep})
 # set compilation flags
 if (UNIX)
   set(${ep}_c_flags "${${ep}_c_flags} -Wall")
-  set(${ep}_cxx_flags "${${ep}_cxx_flags} -Wall")
+  set(${ep}_cxx_flags "${${ep}_cxx_flags} -Wall -Wno-unknown-pragmas")
```
Florent2305: Remember this line modification is here only because of dtk.
It must be removed on 4.1
an equivalent of this fix for Windows would be fine.

src/plugins/legacy/itkDataSHImage/management/vtkSphericalHarmonicSource.cxx
mathildemerle: (my compiler does not like the indentation, it asked for clarity, that's why the "for" loop has now brackets)

superbuild/projects_modules/medInria.cmake
mathildemerle: They are not used as far as i saw, and the diffeomorphic one changed its name

superbuild/projects_modules/medInria.cmake
mathildemerle: oki, i do that as soon as possible

src/plugins/legacy/itkDataSHImage/management/vtkSphericalHarmonicSource.cxx
mathildemerle: it should do:  [number0 number1 ...]

superbuild/projects_modules/medInria.cmake
mathildemerle: @Florent2305 : do you have these pragmas warning messages? If yes, is this parameter working to solve the problem?

Issue #511, opened 2019-11-28
[Ubuntu16] do not compile


mathildemerle:
medInria does not compile on Ubuntu 16. The gcc 5 compiler is not accepted, and C++17 also. 
In order to compile @mjuhoor needed to update the compiler, and change C++17 to C++11.

Florent2305:
4 solutions are available:

- update the compiler to last version of gcc/g++ (My favorite)
- MUSIC fork of medInria set C++ version to Cxx14
- wait few months
- set C++ version to Cxx14 for medInria until april 2020 (The worth for me)


To update Gcc/G++:
`sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y && `
`sudo apt-get update -y && `
`sudo apt-get install gcc-9 g++-9 -y && `
`sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 60 --slave /usr/bin/g++ g++ `
`/usr/bin/g++-9 && `
`sudo update-alternatives --config gcc`

PR #514, opened 2019-12-03
[Ubuntu16] switch back to c++11

mathildemerle:
Fixes https://github.com/medInria/medInria-public/issues/511

Switch C++ used to 11. We do not use c++17, and it's not by default on Ubuntu 16 (at least).

Concerning gcc 7 it can be a documentation solution. I added the [pre-requisites for ubuntu 16](https://github.com/medInria/medInria-public/wiki/A-Quick-Dev-Guide#-pre-requisites-for-compilation) on the doc.

In april 2020, the next Ubuntu LTS is going to be released. At that point, the supported LTS could be 18 and 20, not 16 and 18. On Ubuntu 18 we have c++17.

:m:


Florent2305:
4 solutions are available:

- update the compiler to last version of gcc/g++ (My favorite)
- MUSIC fork of medInria set C++ version to Cxx14
- wait few months
- set C++ version to Cxx14 for medInria until april 2020 (The worth for me)


To update Gcc/G++:
`sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y && 
sudo apt-get update -y && 
sudo apt-get install gcc-9 g++-9 -y && 
sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 60 --slave /usr/bin/g++ g++ /usr/bin/g++-9 && 
sudo update-alternatives --config gcc`

msermesant:
if you want my opinion, can you summarise the advantages / drawbacks of C++11/14/17?
thanks

mathildemerle:
To summarize, there are 2 problems on Ubuntu 16: 
 * there is no gcc7 (just gcc5)
 * there is no c++17 (just 11 or 14)

C++17 is not used in medInria right now, so we can keep the current c++ on the minimal OS versions. 

Ubuntu 16 is kept by MUSIC until April normally. So, it's complicated that the pre-requisites for medInria are not the default ones on the minimum versions handled by medInria or music, and we are working on medInria/music3 right now, so this can't wait i think next April. And this is just on Ubuntu, but i did not used Mac. Perhaps, it would have been better not to switch to gcc7 since it was not the default one on the supported OS. Since it's complicated (maybe? is it?) to switch back to the supported gcc5, we can perhaps ask devs on ubuntu 16 until April to install gcc7 through the non-official-default ppa repositories, but it's a bit annoying.

ocommowi:
There might be other solutions than just reverting back to c++11. Wouldn't it work to use clang rather than g++ on ubuntu 16 ?

That being said, I think it is technically doable to revert back to c++11 though since what we would really use in c++14 or 17 in medinria is not there right now (eg spherical harmonics are not coded on some c++17 compatible compilers, the macos one to mention it).

mathildemerle:
I talked with @paulineMig and @mjuhoor which are on Ubuntu 16. They would like to keep Ubuntu 16 supported even after the release of Ubuntu 20 (for them and/or users of music). They have a lot of compiled soft, and it could be long to recompile everything after an upgrade. They may switch to Ubuntu 20 in april, but they are not sure. We'll talk about that with the team i think. However, they are ok to add pre-requisites specially for gcc7 on Ubuntu 16 to reach the version used on Ubuntu 18. 

Last version of the doc linux https://github.com/medInria/medInria-public/wiki/A-Quick-Dev-Guide#-on-linux 
I can close this PR i think.

PR #520, opened 2019-12-12
[DB] avoid not-found saved data in a new database after restart

mathildemerle:
In order to see the bug:
 * Open medInria -> Settings -> Database -> change the database path to a completely new one -> Save -> Close medInria
 * Open medInria -> Workspace/Browser -> open a data -> permanent saving of the data -> Close medInria
 * Open medInria -> Visualization -> open the data: it does not open.

This PR solves this problem. New databases were not opened well, since the image column does not exist in them, but user_version is not set.

:m:


ocommowi:
Code is sound. Works for me. Merging, thanks

PR #522, opened 2019-12-13
[vtk] write/read metadata in vtk mesh files at export/import

mathildemerle:
From a bug found by @mjuhoor  and @paulineMig, metadata were shown in attributes of a vtk in the Settings.

This PR allows to add metadata in a VTK file.

:m:

### DIFF
--- a/src/plugins/legacy/vtkDataMesh/readers/vtkDataMeshReaderBase.h
+++ b/src/plugins/legacy/vtkDataMesh/readers/vtkDataMeshReaderBase.h
@@ -1,3 +1,15 @@
+/*=========================================================================
+
+ medInria
+
+ Copyright (c) INRIA 2013 - 2019. All rights reserved.
+ See LICENSE.txt for details.
+
+  This software is distributed WITHOUT ANY WARRANTY; without even
+  the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+  PURPOSE.
+
+=========================================================================*/
 #pragma once
 
 #include "vtkDataMeshPluginExport.h"
--- a/src/plugins/legacy/vtkDataMesh/readers/vtkDataMeshReaderBase.h
+++ b/src/plugins/legacy/vtkDataMesh/readers/vtkDataMeshReaderBase.h
@@ -1,0 +1,30 @@
+#pragma once
+
+#include "vtkDataMeshPluginExport.h"
+
+#include <dtkCoreSupport/dtkAbstractDataReader.h>
+
+class vtkMetaDataSet;
+
+class VTKDATAMESHPLUGIN_EXPORT vtkDataMeshReaderBase: public dtkAbstractDataReader
+{
+    Q_OBJECT
+
+public:
+    vtkDataMeshReaderBase();
+
+public slots:
+
+    virtual bool canRead        (const QStringList& paths);
+    virtual bool readInformation(const QStringList& paths);
+    virtual bool read           (const QStringList& paths);
+
+    virtual void setProgress(int value);
+
+    bool extractMetaDataFromFieldData(vtkMetaDataSet* dataSet);
+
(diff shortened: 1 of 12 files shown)


ocommowi:
Do we have an example file for testing that one ?

Thanks

mjuhoor:
There you go, just a left atrium surface mesh
[atria_coarse.vtk.zip](https://github.com/medInria/medInria-public/files/3977819/atria_coarse.vtk.zip)

mjuhoor:
I really believe that 'Field data' should not be available to be displayed in the 'Attributes' combo box of mesh interactors.
The 'Attributes' combo box changes the active (displayed) scalar field on the mesh defined at every point in the case of Point Data, or every cell for Cell Data.
Field data is supposed to hold data that is neither on points nor on cells. Therefore, it makes no sense to show Field Data in the combo box.

fcollot:
There is a parallel discussion related to this PR on a mail thread so to clarify things I'll recall what was said: In MUSIC we need to include the metadata inside the VTK and MHA files so that it can be retrieved when we export the files to another MUSIC installation. Without that, the reimported files have empty metadata. As a result of that functionality, the modified VTKs display the bug described by @mjuhoor, where the metadata tags appear in the attributes combo box.

This PR does not fix that bug. It introduces the code that writes the metadata in the VTK files (which causes the bug to appear). It should not be a problem for VTK files to have field data like this, so I agree with @mjuhoor that the bug should be fixed at the level of what is displayed in the attributes combo box.

On the other hand if you do not want the metadata to be exported as is done in this PR, we will make these changes on our fork since we need them anyways as @juliencastelneau pointed out by mail.

Florent2305:
Ok,

Then, after yours precious responses of Mehdi, Julien, Maxime and Florent.C, I think it's better to wait until after holidays to do something good for all. 
I move this PR to milestone 3.1.1, if a quick solution is find during the begin of January we can re-plan to 3.1.0

fcollot:
The attributes bug is fixed in https://github.com/medInria/medInria-public/pull/526. Note that this fix is still needed for MUSIC, even if the metadata export code presented here is not merged, because our database is full of files that already contain the metadata.

mathildemerle:
@fcollot can i close this PR?

fcollot:
We need this functionality, either here or on our fork, so if this isn't going to be merged here it has to be posted on our core.

mathildemerle:
We thought about this PR with our team, and i added a better description of vtkDataMesh4DWriter and vtkDataMeshWriter export, which export metadata, for instance the patient name, etc. 

I'm going to rename this PR to fit to the final purpose, since https://github.com/medInria/medInria-public/pull/526 has been solving the original display problem.

This functionality can be interesting for medInria, it allows to export and import metadata in vtk mesh files through vtk field data. It simplifies also the creation of future vtkdatamesh writers and readers adding base types. As said above, this functionality is needed for us, so if you do not want it, no problems, we could add it to our fork. An other possibility would be to create 2 kinds of vtk mesh exports, one with metadata, the other without, but these field data are well read by others software.

src/plugins/legacy/vtkDataMesh/readers/vtkDataMeshReaderBase.h
```diff
@@ -0,0 +1,42 @@
...
+
+class VTKDATAMESHPLUGIN_EXPORT vtkDataMeshReaderBase: public dtkAbstractDataReader
+{
+    Q_OBJECT
+
+public:
+    vtkDataMeshReaderBase();
+
+public slots:
+
```
Florent2305: Be careful :
 "virtual bool canRead (const QStringList& paths);" is the old way to specify a function is heritage and already virtual.
new way :
"bool canRead (const QStringList& paths) override;"

mathildemerle:
(i did not finished you comments, i'm going to create a new small PR ^^)

PR #527, opened 2020-01-08
Update to ITK 5.1

Florent2305:
a plus is revert ICU on DCMTK


superbuild/projects_modules/ITK.cmake
```diff
@@ -42,7 +42,7 @@ if (NOT USE_SYSTEM_${ep})
 ## #############################################################################
 
 set(git_url ${GITHUB_PREFIX}InsightSoftwareConsortium/ITK.git)
-set(git_tag v5.0.0)
+set(git_tag origin/release)
```
mathildemerle, line 45: why not v5.1.0 ?

ocommowi: It didn't exist at the time. But it should be nice to do indeed.

Florent2305:
Master moves to ITK5.1.1 by another way

PR #535, opened 2020-01-22
[Dicom import] Fix DICOM import when path contains special characters

### LINKED ISSUE #534: [DICOM Import] Failure to import if path contains special characters
medInria will fail to import DICOMs if the absolute path contains special characters.

1 - create directory named ` isdé_&"&'é""  `
2 - place a dicom inside
3 - import in medInria -> failure

This was previously solved with a brutal `QTextCodec::setCodecForCStrings(QTextCodec::codecForName("UTF-8"));`
but this static function is gone in Qt5

mjuhoor:
Fixes #534 

Problem was due to string conversion from UTF-8 to Latin1.

This could be applied to other readers and/or writers.


mathildemerle:
Beware of the windows system, it could create problem on it ^^

mathildemerle:
@Florent2305 @mjuhoor did you find a way to solve this problem on Linux/Windows/Mac?

Florent2305:
toLocal8Bit instead of toUtf8 is better under Windows systems but is not THE solution because ideogram and other spécial glyph doesn't feet correctly.
I work on a solution with compilation flag for dcmtk, I expect a result for tomorrow.

Florent2305:
This is what was promised and this is what must be delivered.
I done a PR directly on the mjuhoor:fixDicomImport branche of @mjuhoor .
https://github.com/mjuhoor/medInria-public/pull/1

Please test and consider this PR.
If you want test my branche, it's Florent2305:fixDicomImport
Best Flo

PR #552, opened 2020-02-27
Med inria aphp

Florent2305:
Just few changes to compile correctly on Ci mac and windows.
Please, check changes keep the same behavior.


src/plugins/legacy/polygonRoi/polygonEventFilter.cpp
```diff
@@ -31,7 +31,7 @@
 #include <itkComposeImageFilter.h>
 
 polygonEventFilter::polygonEventFilter(medAbstractImageView *view) :
-        medViewEventFilter(), currentView(view), cursorState(CURSORSTATE::MOUSE_EVENT), isRepulsorActivated(false)
+        medViewEventFilter(), currentView(view), cursorState(CURSORSTATE::CS_MOUSE_EVENT), isRepulsorActivated(false)
```
Florent2305: MOUSE_EVENT already exist on Windows environment.
Then I prefix it by 'CS_' for  cursorState to can compile.
Change the name to a better if you want.

src/plugins/legacy/polygonRoi/polygonRoiToolBox.cpp
```diff
@@ -160,6 +160,7 @@ medAbstractData *polygonRoiToolBox::processOutput()
 //        //generateBinaryImage();
 //    }
 //    return m_maskData; // return output data
+    return nullptr;
```
Florent2305: Stub to compile

PR #564, opened 2020-03-23
4D load crash medInria

Florent2305:


mathildemerle:
If possible, is it possible to clarify the title and description of this PR? It will be easier to search this PR if needed in the future :)

src/plugins/legacy/itkDataImage/interactors/medVtkViewItkDataImageInteractor.cpp
```diff
@@ -232,6 +232,8 @@ bool medVtkViewItkDataImageInteractor::SetViewInput(medAbstractData* data, int l
 
 void medVtkViewItkDataImageInteractor::initParameters(medAbstractImageData* data)
 {
+    d->imageData = data; //used among others for 4D
```
mathildemerle: i wonder if it solves also the dicom import crash. This class did not find imageData :o

Florent2305: About 4D Dicom?

mathildemerle: indeed, the one i sent you this morning ^^

Florent2305: From my side it's ok, I mean it's load without crash.
But I don't now this data and maybe it's not render correctly.

Issue #543, opened 2020-02-05
[Copyright] suggestion : easier change


mathildemerle:
This is just a suggestion. I was looking at the copyrights on each file of medInria.

The current copyright:

> /*=========================================================================
> 
>  medInria
> 
>  Copyright (c) INRIA 2013 - 2018. All rights reserved.
>  See LICENSE.txt for details.
>  
>   This software is distributed WITHOUT ANY WARRANTY; without even
>   the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
>   PURPOSE.
> 
> =========================================================================*/

Each time we change a file, we need to check the year. It's complicated to track the right year if we change something and go back to a previous state, it takes time, and it's more difficult to review long PR.

Is it possible to change it to:

> /*=========================================================================
> 
>  medInria
> 
>  Copyright (c) INRIA. All rights reserved.
> 
>  See LICENSE.txt for details in the root of the sources or:
>  https://github.com/medInria/medInria-public/blob/master/LICENSE.txt
>  
>   This software is distributed WITHOUT ANY WARRANTY; without even
>   the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
>   PURPOSE.
> 
> =========================================================================*/

Florent2305:
I think it's a good idea and maybe we can keep a more explicitly clause into the main.cpp

msermesant:
We should keep a date. We can keep just one year, the year of creation of the copyright (2013). There is no need to update.

mathildemerle:
We indeed stopped to use the current year on music, but it was asked on medInria. Can we do that on medInria too?

Something like:

```
/*=========================================================================

 medInria

 Copyright (c) INRIA 2013. All rights reserved.

 See LICENSE.txt for details in the root of the sources or:
 https://github.com/medInria/medInria-public/blob/master/LICENSE.txt

 This software is distributed WITHOUT ANY WARRANTY; without even
 the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
 PURPOSE.

=========================================================================*/
```
Or without the link to the license on Github?

fcollot:
[...]

PR #567, opened 2020-03-24
Sh corrections

Florent2305:
Corrects two problems:

- MaxSH basis was not fitting the one from M Descoteaux PhD, corrected
- SH images were required to be of order less than 6, not a limit anymore
- Reading SH images on my computer was failing on med 3.1.1 future version, corrected

To be merged also to medinria3.1.x branch


src/app/medInria/areas/browser/medBrowserArea.cpp
```diff
@@ -2,7 +2,7 @@
 
  medInria
 
- Copyright (c) INRIA 2013 - 2018. All rights reserved.
+ Copyright (c) INRIA 2013 - 2020. All rights reserved.
```
mathildemerle: it reminds me that it could be interesting to have a final decision for this topic https://github.com/medInria/medInria-public/issues/543 :)

src/app/medInria/areas/browser/medBrowserArea.cpp
```diff
@@ -91,15 +88,17 @@ medBrowserArea::medBrowserArea(QWidget *parent) : QWidget(parent), d(new medBrow
         d->sourceSelectorToolBox->setCurrentTab(1);
     }
 
-    //dataSources
-    foreach (medAbstractDataSource *dataSource, medDataSourceManager::instance()->dataSources())
+    // DataSources
+    for(medAbstractDataSource *dataSource : medDataSourceManager::instance()->dataSources())
```
mathildemerle: this was not supposed to be in 3.1. Did you push the differences between master and 3.1 on 3.1? It could be a problem

PR #573, opened 2020-03-25
[Copyright] update them in CMakeLists with link

mathildemerle:
From the suggestion of copyright in this issue: https://github.com/medInria/medInria-public/issues/543

This PR tests the copyright only in CMakeLists for now. Concerning the old copyright, for instance, we almost never change it since 2020. This new copyright allows to be updated easily.

I updated the main license also.

If ok, i could do it in the other files.

:m:


ocommowi:
Alright for me. Merging then

PR #600, opened 2020-04-10
[ViewContainer] bug with reslice/crop size, and small enhancements -- master

mathildemerle:
Same as https://github.com/medInria/medInria-public/pull/599 but on master branch.

Solve the Cropping/Reslice view size bug.

:m:

### DIFF
--- a/src/layers/legacy/medCoreLegacy/gui/viewContainers/medViewContainer.cpp
+++ b/src/layers/legacy/medCoreLegacy/gui/viewContainers/medViewContainer.cpp
@@ -111,7 +111,7 @@ medViewContainer::medViewContainer(medViewContainerSplitter *parent): QFrame(par
     QVBoxLayout *defaultLayout = new QVBoxLayout(d->defaultWidget);
     defaultLayout->addWidget(defaultLabel);
     defaultLayout->addWidget(openButton);
-    connect(openButton, SIGNAL(clicked()), this, SLOT(openFromSystem()));
+    connect(openButton, SIGNAL(clicked()), this, SLOT(openFromSystem()), Qt::UniqueConnection);
 
     d->menuButton = new QPushButton(this);
     d->menuButton->setIcon(QIcon(":/pixmaps/tools.png"));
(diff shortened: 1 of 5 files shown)


src/layers/legacy/medCoreLegacy/gui/viewContainers/medViewContainer.cpp
```diff
@@ -514,9 +515,11 @@ void medViewContainer::highlight(QString color)
             d->view->mainWindow()->updateGeometry();
             d->view->mainWindow()->update();
         }
-
```
Florent2305, line 517: T'es sûr qu'il faut pas le faire dans tout les cas ?

src/layers/legacy/medCoreLegacy/parameters/medDoubleParameterL.cpp
```diff
@@ -91,16 +91,16 @@ void medDoubleParameterL::setValue(double value)
         }
 
         //  update intern widget
-        this->blockInternWidgetsSignals(true);
```
Florent2305, line 94: Cela change quoi de les avoir déplacer dans l'update du widget?

mathildemerle: this allows to put back the `updateInternWigets` line in `setRange` function which have been removed in a previous PR, but keeping the blocks. If we use `updateInternWigets` we need to block

src/layers/legacy/medCoreLegacy/gui/viewContainers/medViewContainer.cpp
mathildemerle: i am going to test again if you want

mathildemerle: i do not see any differences. In medVtkView we always create mainWindow, and viewWidget is applied on it (`d->mainWindow->setCentralWidget(d->viewWidget);`). I think if you update mainWindow, you update viewWidget. However, if mainWindow is not set (in a new type of view for instance), it would be important to update the viewWidget, in a `else` like here.

Florent2305: ok

src/layers/legacy/medCoreLegacy/parameters/medDoubleParameterL.cpp
Florent2305: ok

PR #601, opened 2020-04-13
minor memory leak : Qlist do not delete it's pointer member, so as do…

nbaraka:
As done it other part of medinria source code, in order to delete a QList which pointers  it is better to call qDeleteAll. (prevent memory leak by calling delete on each pointer)

### DIFF
--- a/src/layers/legacy/medCoreLegacy/views/medAbstractLayeredView.cpp
+++ b/src/layers/legacy/medCoreLegacy/views/medAbstractLayeredView.cpp
@@ -62,6 +62,10 @@ medAbstractLayeredView::~medAbstractLayeredView()
     int c = layersCount()-1;
     for(int i=c; i>=0; i--)
         removeLayer(i);
+
+    qDeleteAll(d->extraNavigators);
+    d->extraNavigators.clear();
+
     delete d;
 }
 


mathildemerle:
@nbaraka is it solving a bug that you found? Or is it enhancement?

mathildemerle:
(if it is a debug, it could be interesting to add it also on the medInria3.1.x branch ^^)

fcollot:
There shouldn't be a memory leak since the navigators are parented to the 'medAbstractLayeredView' instance, and so will be deleted with it.

nbaraka:
> There shouldn't be a memory leak since the navigators are parented to the `medAbstractLayeredView` instance, and so will be deleted with it.
OK I agree but it rely a lot on this QObject feature. 
Futhermore they are a lot of raw pointers that are "exchanged"
between classes and it is difficult to know when you are the owner of 
some memory. 
Using a class from the lib, you have to use it as it used in other 
part of the framework if you do not to fight with a dangling pointer. 
(ex : use carefully the default nullptr for parent)

I notice that **medDataSourceManager** also do not clean its memory 
(but since it is a singleton, it will be instantiated only one time and
it will be deleted when the app quit :)

nbaraka:
So at least it should not have default **parent** parameter in constructor and it could be also more restrictive by not allowing a parent to be a nullptr. What do you think ?

mathildemerle:
Since there is no memory leak: close this PR. If needed, the discussion could continue in an issue

PR #605, opened 2020-04-15
[4D mesh] Fix attribute intensity problem

mjuhoor:
- Fix the displayed attribute's range (called intensity in the code). The min and max values are now computed across the whole sequence.
- For integer arrays, intensity step will be 1 and number of decimals 0.

### DIFF
--- a/src/layers/legacy/medVtkDataMeshBase/vtkMetaDataSetSequence.cxx
+++ b/src/layers/legacy/medVtkDataMeshBase/vtkMetaDataSetSequence.cxx
@@ -496,7 +496,7 @@ void vtkMetaDataSetSequence::UpdateToIndex (unsigned int id)
 //----------------------------------------------------------------------------
 double*vtkMetaDataSetSequence::GetCurrentScalarRange()
 {
-    static double *val = new double[2];
+    double *val = new double[2];
     val[0] = VTK_DOUBLE_MAX;
     val[1] = VTK_DOUBLE_MIN;
 
--- a/src/layers/legacy/medVtkDataMeshBase/vtkMetaDataSetSequence.cxx
+++ b/src/layers/legacy/medVtkDataMeshBase/vtkMetaDataSetSequence.cxx
@@ -722,3 +722,26 @@ void vtkMetaDataSetSequence::ComputeTimesFromDuration()
         this->GetMetaDataSet(i)->SetTime((double) ( i )/ (double)(this->GetNumberOfMetaDataSets() ) );
     }
 }
+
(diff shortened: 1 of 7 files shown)


mathildemerle:
Great! Is this possible to create the same one on master? If the problem is on master too ^^'. In few days (hope) we'll switch full time on the master (4.0) branch, so it could be easier to create just one PR on master

mjuhoor:
> Great! Is this possible to create the same one on master? If the problem is on master too ^^'. In few days (hope) we'll switch full time on the master (4.0) branch, so it could be easier to create just one PR on master

Shouldn't 'master' merge with 'medInria3.1.x' then ? Or cherry-pick ?

mjuhoor:
Ok, I did the cherry-picking to create a new PR, to be merged on  'master'.
However, I will not do this for other PRs. Respectfully, it should be 'core developers' that do this once the PR is merged into medInria3.1.x.

src/layers/legacy/medVtkDataMeshBase/vtkMetaDataSetSequence.cxx
```diff
@@ -722,3 +722,26 @@ void vtkMetaDataSetSequence::ComputeTimesFromDuration()
         this->GetMetaDataSet(i)->SetTime((double) ( i )/ (double)(this->GetNumberOfMetaDataSets() ) );
     }
 }
+
+double* vtkMetaDataSetSequence::GetScalarRange(QString attribute)
+{
+    static double* val = new double[2];
```
Florent2305, line 728: Why val is declared as static pointer ?

Florent2305: I think, but maybe I'm wrong, it is not thread safe et we can have memory leak.
Some other solutions exists:
- std::array (with semantic move it's very clean and easy)
- std::vector  (with semantic move it's very clean and easy)
- use a static 'c' tab

mjuhoor: I agree, static is evil. I just copied this code from medInria 2, and needs to be improved.
I thought about changing the return type as you suggested, or pass the array by parameter.
It would however change the method signature and impact numerous other files, including some external projects.
Maybe in another PR? 
I guess removing the 'static' is already a good thing.

Florent2305: Yes, remove the static it's the minimum to able to merge this PR :)
After, for other solution choose your favorite.

PR #611, opened 2020-04-16
[medInria-3.1.x] Add support for medit .mesh file format

mjuhoor:
- Medit surface and volume mesh are now properly read
- Implements a more robust check for medit file format, as the '.mesh' extension also exists for other file formats (i.e. Biosense Webster CARTO)

### DIFF
--- a/src/layers/legacy/medVtkDataMeshBase/vtkMetaSurfaceMesh.cxx
+++ b/src/layers/legacy/medVtkDataMeshBase/vtkMetaSurfaceMesh.cxx
@@ -335,7 +335,14 @@ unsigned int vtkMetaSurfaceMesh::CanReadFile (const char* filename)
 {
 
   if (vtkMetaSurfaceMesh::IsMeshExtension(vtksys::SystemTools::GetFilenameLastExtension(filename).c_str()))
-    return vtkMetaSurfaceMesh::FILE_IS_MESH;
+  {
+    // medit .mesh format must have 'MeshVersionFormatted' as header
+    if (vtkMetaDataSet::IsMeditFormat(filename))
+    {
+        return vtkMetaSurfaceMesh::FILE_IS_MESH;
+    }
+    return 0;
+  }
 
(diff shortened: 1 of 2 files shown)


src/layers/legacy/medVtkDataMeshBase/vtkMetaSurfaceMesh.cxx
```diff
@@ -335,7 +335,14 @@ unsigned int vtkMetaSurfaceMesh::CanReadFile (const char* filename)
 {
 
   if (vtkMetaSurfaceMesh::IsMeshExtension(vtksys::SystemTools::GetFilenameLastExtension(filename).c_str()))
-    return vtkMetaSurfaceMesh::FILE_IS_MESH;
+  {
+    // medit .mesh format must have 'MeshVersionFormatted' as header
+    if (vtkMetaDataSet::IsMeditFormat(filename))
```
mathildemerle, line 340: (small indentation problem)
I have a question, it's possible that a vtkMetaSurfaceMesh  is not a medit mesh? because it would return 0 here

mjuhoor: I guess you mean: Can 'toto.mesh' be something else than a medit surface mesh ?
Yes, as I said I can be a Carto surface mesh

mathildemerle: oki, i just wanted to be sure, because with a carto or an other non-medit mesh, we do not enter here, and `vtkMetaSurfaceMesh::FILE_IS_MESH;` is not sent

mjuhoor: I see what you mean. There is a confusion between file format (like medit, vtk, Carto...) and filename extension (like .mesh, .vtk, ...)
Here, it should not return vtkMetaSurfaceMesh::FILE_IS_MESH but something like vtkMetaSurfaceMesh::FILE_IS_MEDIT

PR #612, opened 2020-04-17
[medInria-3.1.x] Remove old hardware shading 

Florent2305:
Before this PR HWShading is not used and not accessible because GPU acceleration check box is always inactive


src/plugins/legacy/medVtkFibersData/manager/vtkFibersManager.cxx
```diff
@@ -674,5 +540,4 @@ void vtkFibersManager::Execute()
   this->Callback->GetROIFiberLimiter()->Modified();
   this->Callback->GetFiberLimiter()->Modified();
   this->Callback->GetFiberLimiter()->Update();
-  //m_Callback->Execute(m_BoxWidget, 0, 0);
```
Florent2305, line 677: Note for me, I can remove it

PR #618, opened 2020-04-20
[Q_OS_X11] changes from Qt4->5, code was not used

mathildemerle:
Working on an other problem, i found that Q_OS_X11 and qt_x11_wait_for_window_manager were in QT4 but are not in QT5.

I'm going to create the master one also.

:m:

### DIFF
--- a/src/app/medInria/QSingleApplication/qtsingleapplication.cpp
+++ b/src/app/medInria/QSingleApplication/qtsingleapplication.cpp
@@ -168,47 +168,6 @@ QtSingleApplication::QtSingleApplication(const QString &appId, int &argc, char *
     sysInit(appId);
 }
 
-#if defined(Q_OS_X11)
-/*!
-  Special constructor for X11, ref. the documentation of
-  QApplication's corresponding constructor. The application identifier
-  will be QCoreApplication::applicationFilePath(). \a dpy, \a visual,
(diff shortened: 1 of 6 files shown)


src/app/medInria/QSingleApplication/qtsingleapplication.cpp
```diff
@@ -168,47 +168,6 @@ QtSingleApplication::QtSingleApplication(const QString &appId, int &argc, char *
...
 }
 
-#if defined(Q_OS_X11)
-/*!
-  Special constructor for X11, ref. the documentation of
-  QApplication's corresponding constructor. The application identifier
-  will be QCoreApplication::applicationFilePath(). \a dpy, \a visual,
-  and \a cmap are passed on to the QApplication constructor.
-*/
-QtSingleApplication::QtSingleApplication(Display* dpy, Qt::HANDLE visual, Qt::HANDLE cmap)
```
mathildemerle, line 178: removed because these functions are not used anymore, and "Display" is not known

mathildemerle: Used: https://doc.qt.io/archives/qt-4.8/qapplication.html but not used anymore in https://doc.qt.io/qt-5/qapplication.html

PR #619, opened 2020-04-20
add some test before dereferencing pointers

nbaraka:
Add tests when getting a pointer from a function that could return a nullptr.

### DIFF
--- a/src/layers/legacy/medVtkInria/vtkImageView/vtkImageView2D.cxx
+++ b/src/layers/legacy/medVtkInria/vtkImageView/vtkImageView2D.cxx
@@ -241,12 +241,13 @@ vtkMTimeType vtkImageView2D::GetMTime()
 //----------------------------------------------------------------------------
 void vtkImageView2D::GetSliceRange(int &min, int &max) const
 {
-    if (this->GetMedVtkImageInfo()->initialized)
+    if (this->GetMedVtkImageInfo() && this->GetMedVtkImageInfo()->initialized)
     {
         min = 0;
         max = 0;
         
-        this->Get2DDisplayMapperInputAlgorithm()->UpdateInformation();
+        if(this->Get2DDisplayMapperInputAlgorithm())
+            this->Get2DDisplayMapperInputAlgorithm()->UpdateInformation();
         auto pAlgo = this->Get2DDisplayMapperInputAlgorithm();
         if (pAlgo != nullptr)
         {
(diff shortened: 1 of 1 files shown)


src/layers/legacy/medVtkInria/vtkImageView/vtkImageView2D.cxx
```diff
@@ -241,12 +241,13 @@ vtkMTimeType vtkImageView2D::GetMTime()
...
 void vtkImageView2D::GetSliceRange(int &min, int &max) const
 {
-    if (this->GetMedVtkImageInfo()->initialized)
+    if (this->GetMedVtkImageInfo() && this->GetMedVtkImageInfo()->initialized)
     {
         min = 0;
         max = 0;
         
-        this->Get2DDisplayMapperInputAlgorithm()->UpdateInformation();
+        if(this->Get2DDisplayMapperInputAlgorithm())
```
mathildemerle, line 249: can you add brackets?

src/layers/legacy/medVtkInria/vtkImageView/vtkImageView2D.cxx
```diff
@@ -640,7 +641,8 @@ void vtkImageView2D::SetViewOrientation(int orientation)
 void vtkImageView2D::SetOrientationMatrix (vtkMatrix4x4* matrix)
 {
   this->Superclass::SetOrientationMatrix (matrix);
-  this->GetImage2DDisplayForLayer(GetFirstLayer())->GetImageActor()->SetUserMatrix (this->OrientationMatrix);
+    if(this->GetImage2DDisplayForLayer(GetFirstLayer()))
```
mathildemerle, line 644: {} ?

src/layers/legacy/medVtkInria/vtkImageView/vtkImageView2D.cxx
```diff
@@ -1743,16 +1745,19 @@ void vtkImageView2D::SetInput(vtkAlgorithmOutput* pi_poVtkAlgoOutput, vtkMatrix4
...
     this->AddLayer(layer);
     pi_poVtkAlgoOutput = this->ResliceImageToInput(pi_poVtkAlgoOutput, matrix);
-
     vtkImage2DDisplay *imageDisplay = this->GetImage2DDisplayForLayer(layer);
-    imageDisplay->SetInputProducer(pi_poVtkAlgoOutput);
-    imageDisplay->SetInputData(static_cast<vtkImageAlgorithm*>(pi_poVtkAlgoOutput->GetProducer())->GetOutput());
-    imageDisplay->GetImageActor()->SetUserMatrix (this->OrientationMatrix);
-
-    this->SetColorRange(imageDisplay->GetMedVtkImageInfo()->scalarRange, layer);
+    if (imageDisplay)
```
mathildemerle, line 1751: is imageDisplay could be null? If yes, it should be a problem higher perhaps?

nbaraka: I did not understood. image is has be retrieved by GetImage2DDisplayForLayer which could return a null ptr.

mathildemerle: I mean, GetImage2DDisplayForLayer is set in these files, and if it's null here, we have an algorithmic problem i think, and could be solved. The test on GetImage2DDisplayForLayer or GetImage3DDisplayForLayer is not done in vtkImageView3D neither. That's why i was asking if you saw a data or a process leading to GetImage2DDisplayForLayer being null, and if it can be solved :)

Issue #593, opened 2020-04-06
[Tests] 3.1.1 and master - run twice process in Filtering - bug


mathildemerle:
In Filtering workspace:
 * for instance, open Add a constant,
 * run twice the process -> moving the mouse on the image and tlbx makes them crazy

It's the same with a lot of Filtering toolboxes

mathildemerle:
this problem can be found on master too

PR #631, opened 2020-04-23
[FilteringWorkspaceL] remove signals duplication

mathildemerle:
Working on https://github.com/medInria/medInria-public/issues/593 i found some duplication of signals. 

:m:

### DIFF
--- a/src/plugins/legacy/itkFilters/itkFiltersToolBox.cpp
+++ b/src/plugins/legacy/itkFilters/itkFiltersToolBox.cpp
@@ -496,6 +496,7 @@ void itkFiltersToolBox::clear()
     d->maxColor = Qt::white;
     d->thresholdColor = Qt::black;
     d->minValueImage = d->maxValueImage = 0.;
+    d->process = nullptr;
 }
 
 void itkFiltersToolBox::update()
--- a/src/plugins/legacy/medFilteringWorkspaceL/medFilteringWorkspaceL.cpp
+++ b/src/plugins/legacy/medFilteringWorkspaceL/medFilteringWorkspaceL.cpp
@@ -35,8 +35,6 @@ public:
 medFilteringWorkspaceL::medFilteringWorkspaceL(QWidget *parent)
     : medSelectorWorkspace (parent, staticName()), d(new medFilteringWorkspaceLPrivate)
 {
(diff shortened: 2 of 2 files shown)


src/plugins/legacy/medFilteringWorkspaceL/medFilteringWorkspaceL.cpp
```diff
@@ -35,8 +35,6 @@ class medFilteringWorkspaceLPrivate
 medFilteringWorkspaceL::medFilteringWorkspaceL(QWidget *parent)
     : medSelectorWorkspace (parent, staticName()), d(new medFilteringWorkspaceLPrivate)
 {
-    connect(this->tabbedViewContainers(), SIGNAL(containersSelectedChanged()),
```
Florent2305, line 38: I don't understand why it need to be removed

mathildemerle: inputChanged is already called in medFilteringWorkspaceL::changeToolBoxInput() below

mathildemerle: And this line emits inputChanged even if the output image is changed. We do not want that. That's why in medFilteringWorkspaceL::setupTabbedViewContainer() we choose what we want to do with either the input or the output container

Florent2305: ok

src/plugins/legacy/itkFilters/itkFiltersToolBox.cpp
```diff
@@ -496,6 +496,7 @@ void itkFiltersToolBox::clear()
     d->maxColor = Qt::white;
     d->thresholdColor = Qt::black;
     d->minValueImage = d->maxValueImage = 0.;
+    d->process = nullptr;
```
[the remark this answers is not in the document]
mathildemerle: this is a smartpointer

src/plugins/legacy/medFilteringWorkspaceL/medFilteringWorkspaceL.cpp
Florent2305: Indeed :)

PR #637, opened 2020-04-27
Correction of very low windowLevel images (range < 10^-2)

### LINKED ISSUE #621: Very Low Range  image not usable (eg 0 to 0.005)
quand on droppe une image avec un range tres faible (genre image d'adc entre 0 et 0.005 en gros), les valeurs dans les min max intensities boxes sont toutes les deux a zero et non changeables

From Oliver

Florent2305:
Fix #621 Very Low Range image not usable (eg 0 to 0.005)
On 3.1.x

### DIFF
--- a/src/plugins/legacy/itkDataImage/interactors/medVtkViewItkDataImageInteractor.cpp
+++ b/src/plugins/legacy/itkDataImage/interactors/medVtkViewItkDataImageInteractor.cpp
@@ -337,7 +337,7 @@ void medVtkViewItkDataImageInteractor::initWindowLevelParameters(double *range)
     if(d->isFloatImage)
     {
         int iDecimalCount = 2;
-        if(std::fabsl(d->intensityStep)<1)
+        if(d->intensityStep<1)
         {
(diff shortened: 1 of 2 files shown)


src/plugins/legacy/itkDataImage/interactors/medVtkViewItkDataImageInteractor.cpp
```diff
@@ -336,8 +336,13 @@ void medVtkViewItkDataImageInteractor::initWindowLevelParameters(double *range)
 
     if(d->isFloatImage)
     {
-        d->minIntensityParameter->setDecimals(2);
-        d->maxIntensityParameter->setDecimals(2);
+        int iDecimalCount = 2;
+        if(std::fabsl(d->intensityStep)<1)
```
mathildemerle, line 340: so, the absolute value was needed?

Florent2305: No :)

Florent2305: done and push

PR #639, opened 2020-04-27
[medInria-4.0] Add support for medit .mesh file format

Florent2305:
- Medit surface and volume mesh are now properly read
- Implements a more robust check for medit file format, as the '.mesh' extension also exists for other file formats (i.e. Biosense Webster CARTO)

Cf   [#611](https://github.com/medInria/medInria-public/pull/611)


mathildemerle:
I did the same PR on master with the @mjuhoor he only did on 3.1.x https://github.com/medInria/medInria-public/pull/634 but yours seems that have some code in addition. It's complicated XD.

Is it possible to add a description to this PR?

Florent2305:
Doublon avec celle de @mathildemerle #634

Issue #595, opened 2020-04-07
Layer group 1 appears in link view properties


ocommowi:
A minor bug here. There used to be a predefined group in each of the two categories on the right column : link view properties and link layer properties. Those two were named view group 1 and layer group 1. 

In medinria 3.0, layer group 1 actually appears in link view properties, leaving link layer properties an empty list.

We can go around this bug but solving it could be nice.

PR #641, opened 2020-04-28
[Menu Links] Layer group 1 appears in link view properties

mathildemerle:
Fix https://github.com/medInria/medInria-public/issues/595

The layer Group is in the layer menu, and the view Group is in the view menu. It's solved by the last line of this PR.

I removed a duplicated computation also in loops.

I'm going to do the master PR as well.

:m:

### DIFF
--- a/src/layers/legacy/medCoreLegacy/gui/medAbstractWorkspaceLegacy.cpp
+++ b/src/layers/legacy/medCoreLegacy/gui/medAbstractWorkspaceLegacy.cpp
@@ -697,16 +697,18 @@ QWidget* medAbstractWorkspaceLegacy::buildViewLinkMenu()
 
     // build a stringList witth all the param names
     foreach(medAbstractParameterL* viewParam, viewsParams)
+    {
         paramNames << viewParam->name();
+    }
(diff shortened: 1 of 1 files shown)


src/layers/legacy/medCoreLegacy/gui/medAbstractWorkspaceLegacy.cpp
```diff
@@ -697,16 +697,18 @@ QWidget* medAbstractWorkspaceLegacy::buildViewLinkMenu()
 
     // build a stringList witth all the param names
     foreach(medAbstractParameterL* viewParam, viewsParams)
```
Florent2305, line 699: for :
not needed but... ;)

mathildemerle: that's not in this PR on 3.1.1 but it's already done on the master

Florent2305: ;)

PR #643, opened 2020-04-28
remove QGLWidget used to create a valide context since it is deprecated class

nbaraka:
A way to call glGetString that do not need to construct a QGLWidget. 
Also increase format buffer size.


mathildemerle:
No crash on my side. Do i need to test specific things?

My computer:
IHULUX002-system

Issue #645, opened 2020-04-29
[Ubuntu 18.04] compilation error in std ceill etc


mathildemerle:
```
/home/mathildemerle/Dev/MUSIC3/medInria-public/src/plugins/legacy/itkDataImage/interactors/medVtkViewItkDataImageInteractor.cpp: In member function ‘void medVtkViewItkDataImageInteractor::initWindowLevelParameters(double*)’:
/home/mathildemerle/Dev/MUSIC3/medInria-public/src/plugins/legacy/itkDataImage/interactors/medVtkViewItkDataImageInteractor.cpp:342:38: error: ‘ceill’ is not a member of ‘std’
             iDecimalCount = 1 + std::ceill(std::fabsl(std::log10l(d->intensityStep)));
                                      ^~~~~
/home/mathildemerle/Dev/MUSIC3/medInria-public/src/plugins/legacy/itkDataImage/interactors/medVtkViewItkDataImageInteractor.cpp:342:38: note: suggested alternative: ‘ceil’
             iDecimalCount = 1 + std::ceill(std::fabsl(std::log10l(d->intensityStep)));
                                      ^~~~~
                                      ceil
/home/mathildemerle/Dev/MUSIC3/medInria-public/src/plugins/legacy/itkDataImage/interactors/medVtkViewItkDataImageInteractor.cpp:342:49: error: ‘fabsl’ is not a member of ‘std’
             iDecimalCount = 1 + std::ceill(std::fabsl(std::log10l(d->intensityStep)));
                                                 ^~~~~
/home/mathildemerle/Dev/MUSIC3/medInria-public/src/plugins/legacy/itkDataImage/interactors/medVtkViewItkDataImageInteractor.cpp:342:49: note: suggested alternative: ‘fabs’
             iDecimalCount = 1 + std::ceill(std::fabsl(std::log10l(d->intensityStep)));
                                                 ^~~~~
                                                 fabs
/home/mathildemerle/Dev/MUSIC3/medInria-public/src/plugins/legacy/itkDataImage/interactors/medVtkViewItkDataImageInteractor.cpp:342:60: error: ‘log10l’ is not a member of ‘std’
             iDecimalCount = 1 + std::ceill(std::fabsl(std::log10l(d->intensityStep)));
                                                            ^~~~~~
[...]

PR #646, opened 2020-04-29
[Ubuntu18.04] compilation ceill fabsl log10l

mathildemerle:
Solves https://github.com/medInria/medInria-public/issues/645

On Ubuntu at least, no ceill, fabsl, log10l.

:m:


Florent2305:
It work well on OSX and Windows and it's in C++ 11 official norme cf https://en.cppreference.com/w/cpp/numeric/math/ceil
But I'm ok if it solve this problem on 18.04

PR #652, opened 2020-04-29
memory leak, the caller is the owner 

nbaraka:
Should find a way to make it more obvious (in vtkItkConversion)  to the developer  with a specific pattern in the function name or to use smart pointer (better sol). 

I do not have 4D data. if someone can send me a small data I will be happy to test it. 

I have check that in vtkImageView a deep copy is done the matrix and here it becomes a smart pointer.

### DIFF
--- a/src/layers/legacy/medImageIO/vtkItkConversion.tpp
+++ b/src/layers/legacy/medImageIO/vtkItkConversion.tpp
@@ -302,7 +302,7 @@ double * vtkItkConversion<volumeType, imageDim>::getCurrentScalarRange()
 
     if (m_ItkInputImage.IsNotNull() || m_ItkInputImage4D.IsNotNull())
     {
-        dResScalarRange = new double[2]; //FloTODO check potential memory leak
+        dResScalarRange = new double[2]; //FloTODO check potential memory leak // neb : the caller is the owner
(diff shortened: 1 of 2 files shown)


fcollot:
An easy fix (without needing to change the data type) would be have the caller provide the output table as an argument.

src/layers/legacy/medImageIO/vtkItkConversion.tpp
```diff
@@ -302,7 +302,7 @@ double * vtkItkConversion<volumeType, imageDim>::getCurrentScalarRange()
 
     if (m_ItkInputImage.IsNotNull() || m_ItkInputImage4D.IsNotNull())
     {
-        dResScalarRange = new double[2]; //FloTODO check potential memory leak
+        dResScalarRange = new double[2]; //FloTODO check potential memory leak // neb : the caller is the owner
```
mathildemerle, line 305:
```suggestion
        dResScalarRange = new double[2]; //FloTODO check potential memory leak
```

mathildemerle: (seen with @Florent2305 , need to be fully checked)

mathildemerle (approved):

Florent2305 (approved):

Issue #657, opened 2020-04-30
2 Volume VR & LUT 


nbaraka:
When 2 volumes are opened in the VR mode, one of layer is not responding to the LUT change. 
A change in lut of this layer os only taken into account when other layer's LUT is changed.

PR #661, opened 2020-05-01
improve #326 and solve  #657 

nbaraka:
Major bug : missing call to internalUpdate after a new layer is added. 

Some minor changes to reduce the number of function calls, remove unecessary line, improve the lut management by the scalar range : related #326 #657 ,2 images with big diff in their scalar ranges

In order to overcome the 4 layers limit I would like to change how the layer are managed in the volume rendering. 
With less perf. constraint, it might be possible to improve the rendering of the diff layers and to make the code easier to understand. Now layer index are related index of RGBV (difficult to understand but it seems to be a nice trick)

### DIFF
--- a/src/layers/legacy/medVtkInria/vtkImageView/vtkImageView3D.cxx
+++ b/src/layers/legacy/medVtkInria/vtkImageView/vtkImageView3D.cxx
@@ -46,77 +46,77 @@ vtkStandardNewMacro(vtkImageView3D)
 //----------------------------------------------------------------------------
 vtkImageView3D::vtkImageView3D()
 {
-  this->VolumeProperty  = vtkVolumeProperty::New();
-  this->VolumeActor     = vtkVolume::New();
+  VolumeProperty  = vtkVolumeProperty::New();
+  VolumeActor     = vtkVolume::New();
 
-  this->VolumeMapper = vtkSmartVolumeMapper::New();
+  VolumeMapper = vtkSmartVolumeMapper::New();
 
-  this->Callback    = vtkImageView3DCroppingBoxCallback::New();
-  this->BoxWidget   = vtkOrientedBoxWidget::New();
-  this->PlaneWidget = vtkPlaneWidget::New();
-  this->Marker      = vtkOrientationMarkerWidget::New();
-  this->Cube        = vtkAnnotatedCubeActor::New();
+  Callback    = vtkImageView3DCroppingBoxCallback::New();
+  BoxWidget   = vtkOrientedBoxWidget::New();
+  PlaneWidget = vtkPlaneWidget::New();
+  Marker      = vtkOrientationMarkerWidget::New();
+  Cube        = vtkAnnotatedCubeActor::New();
 
-  this->PlanarWindowLevelX = vtkSmartPointer<vtkImageMapToColors>::New();
-  this->PlanarWindowLevelY = vtkSmartPointer<vtkImageMapToColors>::New();
-  this->PlanarWindowLevelZ = vtkSmartPointer<vtkImageMapToColors>::New();
-  this->ActorX = vtkImageActor::New();
-  this->ActorY = vtkImageActor::New();
-  this->ActorZ = vtkImageActor::New();
+  PlanarWindowLevelX = vtkSmartPointer<vtkImageMapToColors>::New();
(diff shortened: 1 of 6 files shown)


mathildemerle:
@nbaraka i saw a debug in your PR, some rendering improvements tests, and some coding styles changes. Is it possible to focus on the debug if you find one and create a PR? And concerning the removal of "this" i'm not sure that medInria wants to remove the "this" everywhere (and we do not remove them in music). Is it possible not to change too much various thing in one PR, specially the "this"? ^^

nbaraka:
Ready to be merged. Changes have been done only to improve readability in order to solve the bug. "This" might be necessary only for the call of function which have a using def (ie 2) but not to acess data members or call other func.

fcollot:
I personally never use the `this` pointer unless necessary, as @nbaraka pointed out, so all my code doesn't have it. I don't think we should make it a "rule" to have it. There is already a lot of unnecessary code in medInria/MUSIC.

Florent2305:
Personally I don't like 'this->' everywhere.
But it's true, for the moment we don't have a rule for medInria about it.
3 possible cases  :
 - 'this' is mandatory
 - 'this' is at the choice of developer
 - 'this' is by default forbidden

For me I prefer when the developer have the choice then my preference ranking is 2, 3, 1.

But I'm agree with Mathilde to don't personalize coding style during a debug PR.

fcollot:
Yes I agree it should be a developer choice.
I also agree that PRs should not mix functional and style changes, and this has unfortunately been done a lot in the past, so (to take a french expression) we can't really throw the first stone at Nordine...

src/layers/legacy/medVtkInria/vtkImageView/vtkImageView3D.cxx
```diff
@@ -667,103 +662,90 @@ void vtkImageView3D::SetTransferFunctions (vtkColorTransferFunction * color,
...
-    {
-        vtkImage3DDisplay *imageDisplay = GetImage3DDisplayForLayer(layer);
-        if (imageDisplay)
-        {
-            dfRes = imageDisplay->GetOpacity();
-        }
-    }
-    return dfRes;
+    vtkImage3DDisplay * imageDisplay;
+    return (imageDisplay=GetImage3DDisplayForLayer(layer))? imageDisplay->GetOpacity() : 0.0;
```
mathildemerle, line 676: the previous code was more clear. Is it possible to put it back, except if there is a debug in it?

src/layers/legacy/medVtkInria/vtkImageView/vtkImageView3D.cxx
```diff
@@ -1044,31 +1020,24 @@ int vtkImageView3D::GetNumberOfLayers() const
 {
     // I don't really know why, but LayerInfoVec size is set to 1 at initialization time,
     // so we need one more check to know the real number of layer
-    if( this->LayerInfoVec.size() == 1)
+    if( LayerInfoVec.size() == 1)
```
mathildemerle, line 1023: To be clearer, and avoid 2 returns:
```c++
 int result = static_cast<int>(LayerInfoVec.size());
 if( LayerInfoVec.size() == 1)
 {
      result = (LayerInfoVec.at(0).ImageDisplay->GetMedVtkImageInfo())? 1 : 0;
 }
 return result;
```

src/layers/legacy/medVtkInria/vtkImageView/vtkImageView3D.cxx
```diff
@@ -1088,43 +1057,52 @@ void vtkImageView3D::RemoveLayer (int layer)
...
-
-//! Get layer specific info
 vtkImage3DDisplay * vtkImageView3D::GetImage3DDisplayForLayer( int layer ) const
 {
-  if ( layer > -1 && (unsigned int)layer < this->LayerInfoVec.size())
-  {
-      return this->LayerInfoVec.at(layer).ImageDisplay;
-  }
-  else return nullptr;
+   return  ( layer > -1 && (unsigned int)layer < LayerInfoVec.size())? LayerInfoVec.at(layer).ImageDisplay : nullptr;
```
mathildemerle, line 1105: it's less readable, and long, is it possible to switch back?

mathildemerle:
Great :). Is it possible to create a PR on medInria3.1.x with just the debug?

PR #667, opened 2020-05-05
minor memory leak in voiCutterToolBox

nbaraka:
missing delete on two QList. (done some tests : amazing toolbox) and on destructor.


src/plugins/legacy/voiCutter/voiCutterToolBox.cpp
```diff
@@ -758,6 +760,17 @@ void voiCutterToolBox::definePolygonsImage(std::vector<vtkVector2i> polygonPoint
     {
         cutThroughImage<itk::Image<double,3> >(RoiList,RoiPointList,stackMax,stackOrientation,m);
     }
+
+    for(auto poly : *RoiList)
```
mathildemerle, line 764: can you put brackets on these for?

Florent2305 (approved):
ok, but style is not the best

mathildemerle (approved):

PR #668, opened 2020-05-05
minor leak 

nbaraka:
missing delete on save 
modernize by override and auto


nbaraka:
Close this one. It introduces random effect on save result after switch.

PR #671, opened 2020-05-06
[PolygonROI] - Menu predefined labels according to speciality

juliencastelneau:
1. Add ComboBox with specialities 
      - For now, only default and prostate
      - We could add cardiology and brain for instance
      - It is hardcoded actually. 
      - In a future PR, I would like to add a json configuration file to allow users adding new speciality dynamically

2. When a speciality is selected, a listWidget appears with a list of predefined label name according to speciality
      - Ability to choose the order (drag and drop)
      - ability to add label name
      - the choice must be validated with the "set predefined label" button

### DIFF
--- a/src/plugins/legacy/polygonRoi/polygonRoiToolBox.cpp
+++ b/src/plugins/legacy/polygonRoi/polygonRoiToolBox.cpp
@@ -31,10 +31,19 @@
 
 const char *polygonRoiToolBox::generateBinaryImageButtonName = "generateBinaryImageButton";
 
-
 polygonRoiToolBox::polygonRoiToolBox(QWidget *parent ) :
     medAbstractSelectableToolBox(parent), viewEventFilter(nullptr)
 {
+    colorsList = QList<QColor>({
+        Qt::green,
+        Qt::blue,
+        Qt::yellow,
+        Qt::darkMagenta,
+        Qt::cyan,
+        Qt::gray,
+        Qt::white,
+        Qt::black
+    });
     QWidget *displayWidget = new QWidget(this);
     this->addWidget(displayWidget);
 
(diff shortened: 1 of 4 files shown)


src/plugins/legacy/polygonRoi/polygonRoiToolBox.cpp
```diff
@@ -59,16 +68,54 @@ polygonRoiToolBox::polygonRoiToolBox(QWidget *parent ) :
...
+    ButtonLayout2->addStretch(0);
+    ButtonLayout2->setContentsMargins(0,0,0,0);
+    ButtonLayout2->setSpacing(0);
+    QHBoxLayout *ButtonLayout3 = new QHBoxLayout();
+    ButtonLayout3->addWidget(plusButton, Qt::AlignLeft);
+    ButtonLayout3->addStretch(2);
+    ButtonLayout3->addWidget(applyButton, Qt::AlignRight);
+    ButtonLayout1->addLayout(ButtonLayout2);
+    ButtonLayout1->addLayout(ButtonLayout3);
+
```
mathildemerle, line 113: it's really negligible, but you can remove an empty line here

src/plugins/legacy/polygonRoi/polygonRoiToolBox.cpp
```diff
@@ -59,16 +68,54 @@ polygonRoiToolBox::polygonRoiToolBox(QWidget *parent ) :
...
 
     QHBoxLayout *ButtonLayout0 = new QHBoxLayout();
     layout->addLayout( ButtonLayout0 );
     ButtonLayout0->addWidget(addNewCurve);
 
-    QHBoxLayout *ButtonLayout1 = new QHBoxLayout();
-    layout->addLayout( ButtonLayout1 );
-    ButtonLayout1->addWidget(interpolate);
-    ButtonLayout1->addWidget(repulsorTool);
+    QVBoxLayout *ButtonLayout1 = new QVBoxLayout();
```
mathildemerle, line 99: i'm not a big fan of "button1", "button2", etc names for widget. And as you saw, if we add a new group, we need to change every following number. I would prefer something like "layoutSpecialities" or "structureListLayout", etc ^^

src/plugins/legacy/polygonRoi/polygonRoiToolBox.cpp
```diff
@@ -611,6 +664,95 @@ void polygonRoiToolBox::saveBinaryImage()
     viewEventFilter->saveMask();
 }
 
+void polygonRoiToolBox::addLabelName()
+{
+    if ( structureList->count()==colorsList.size())
+    {
+        displayMessageError("Unable to create more label. Maximum size achieved");
+        plusButton->setDisabled(true);
+        return;
```
mathildemerle, line 673: in order to avoid `return` in functions when it could be avoided, you can change this `return` to an `else` after this `if`. The code would be clearer :)

juliencastelneau:
Yes, I will apply the Mathilde's comment in the next PR. 
I think There will be a new one soon

Issue #669, opened 2020-05-05
RESLICER crash with some parameters 


nbaraka:
Changing the parameters spacing create a crash. 
But i got an idea on this.

nbaraka:
@mathildemerle : I think that in **applyResamplingPix** ,  resamplePr->setParameter(outputSpacing[0], 0) should be  resamplePr->setParameter(outputSpacing[0], 3) 
and ... 
Do you agree ?

mathildemerle:
```c++
    resamplePr->setParameter(outputSpacing[0], 3);
    resamplePr->setParameter(outputSpacing[1], 4);
    resamplePr->setParameter(outputSpacing[2], 5);
```
This? It could be indeed, as far as i see, but i did not test :)

nbaraka:
ok thanks I will give it a try.

PR #672, opened 2020-05-06
solve #669 in parameters retrieving 

nbaraka:
When there is no parameters changes when switching between **space** and **dimension** then the parameters are not updated : **space** parameters are used in **dimension**.


mathildemerle:
Great! Is it possible to create the same PR on medInria3.1.x?
(and, can you join our Mattermost servers?)

PR #679, opened 2020-05-07
polygonRoiToolBox mem leak 

nbaraka:
polygonRoiToolBox mem leak.


nbaraka:
Few changes, 
To close  it  is not fair for the work to search for this mem leak. 
I think It brings benefit : mem consumpt, code history.

PR #680, opened 2020-05-07
[medInria3.1.x]solve #669 in parameters retrieving

Florent2305:
port of nbaraka:spacingReslicer
equivalent to #672 in parameters retrieving
RESLICER crash with some parameters #669

### DIFF
--- a/src/plugins/legacy/reformat/medResliceViewer.h
+++ b/src/plugins/legacy/reformat/medResliceViewer.h
@@ -1,14 +1,14 @@
 #pragma once
 /*=========================================================================
 
-medInria
+  medInria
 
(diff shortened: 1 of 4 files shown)


src/plugins/legacy/reformat/medResliceViewer.h
```diff
@@ -1,16 +1,16 @@
+#pragma once
 /*=========================================================================
 
- medInria
+medInria
 
- Copyright (c) INRIA 2019. All rights reserved.
- See LICENSE.txt for details.
+Copyright (c) INRIA 2013 - 2020. All rights reserved.
```
mathildemerle, line 6: there are some problems with the copyrights in this PR

PR #647, opened 2020-04-29
Glyph3d  to Glyph3dMapper 

nbaraka:
changes Glyph3d to Glyph3dMapper in order to benefit from GPU perf. 

Other benefits : 
     
- The rendering seems meaningfull with a test data. 
- no crash in 3D when changing orientation 

Disadvantages : 

- arrow are smaller in 3D

Issue #640, opened 2020-04-27
3D mode crash for Image with vector data or tensor when data is 2D:   "itkDataImageVectorFloat3


nbaraka:
After the import of an image with vector data (3), there is a crash during the change of orientation in the 3d mode. 
(the image is result of computing  gradient of a dicom and saving it in mha file) 

@Florent2305 this bug impact also tensors glyph

mathildemerle:
Hi @nbaraka , do you have a data creating the crash we could test?

mathildemerle:
I have the crash on Ubuntu, on master branch in `vtkInteractorStyleTrackballCamera2::Rotate()` line 102 `rwi->Render();`

Thx for the test data.

nbaraka:
You're welcome. 
I got the same error. I was doing some test to replace vtkGlyph with vtkGlyphMapper which according to https://blog.kitware.com/wp-content/uploads/2016/01/kitware_quarterly1011.pdf has better perf.

mathildemerle:
Does this PR https://github.com/medInria/medInria-public/pull/647 solve this issue @nbaraka ?

PR #682, opened 2020-05-07
Glyph3d mapper same as #647 after rebase

nbaraka:
Glyph3d mapper same as #647 after rebase

### DIFF
--- a/src/plugins/legacy/itkDataImage/interactors/medVtkViewItkVectorFieldInteractor.cpp
+++ b/src/plugins/legacy/itkDataImage/interactors/medVtkViewItkVectorFieldInteractor.cpp
@@ -464,7 +464,20 @@ void medVtkViewItkVectorFieldInteractor::updateSlicingParam()
 
 void medVtkViewItkVectorFieldInteractor::updatePlaneVisibility()
 {
-    // this is a temp sol : not optimal,without this the screen is black
-    this->update();
-    d->manager->SetGlyphScale(1.0);
+    markMapperModified();
 }
+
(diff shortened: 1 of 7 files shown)


mathildemerle:
NB: from old PR closed https://github.com/medInria/medInria-public/pull/647

> @mathildemerle yes it solve #640 tbut he rendering is less smart in 3D. But It do not think that it is relevant to watch it in 3D. (but at least there is no crash)

msermesant:
I use regularly 3D directions in 3D (through the tensor orientations in the diffusion workspace) to check the transmural changes of cardiac fibres, so the rendering of vectors is also relevant in 3D.

Florent2305:
Indeed, at Rennes for exemple 3D is also regularly used for brain tensors applications.

nbaraka:
The issue might not be from the this solution but from the 3D view. I have just opened a mesh (created from a mask) but that is thin (like this vector field) and it is difficult to watch it on 3D without another loaded 3D data. Maybe if you load this 2 vector field after the load of a full 3D data, you don't have this pb.

mathildemerle:
I tried this Pull Request with the "Gradient.mha" data you sent us @nbaraka on 28/04/2020, there is no crash switching in 3D, but there is no displayed data in the view. If you rotate the view, you can see a thin layer, but i don't think your data should looked like that. And if you continue the rotation, medInria crash.
I put this PR on invalid for now, just to be sure not to merge it yet :)

Florent2305:
The original issue is solved in #706

PR #684, opened 2020-05-07
itkDataTensorImageVtkViewInteractor & itkDataSHImageVtkViewInteractor minor mem leak and override

nbaraka:
minor mem leak and override (that is more secure than virtual : your add one more constraint, the function has to be declared in the base class and to be virtual, if a change occured in the base class API than it will be reported as an error during compilation)


Florent2305 (changes requested):
**d->manager->Delete();
d->orientationMatrix->Delete();**
are not the good way
better way consite to centralyze the deletion into a ~itkDataSHImageVtkViewInteractorPrivate like:
~itkDataSHImageVtkViewInteractorPrivate ()
{
    d->manager->Delete();
    d->orientationMatrix->Delete();
}
And finally the best way is to delacre **manager** and **orientationMatrix** as **vtkSmartPointer**

PR #691, opened 2020-05-12
Fix #581 once min/max intensities in 3D V4.0

### LINKED ISSUE #581: [Tests] 3.1.1 and master -> once min/max intensities in 3D, it's kept
On 3.1.1 and master branches:
 * open a data, change the min/max intensities to something "visual"
 * switch to 3D -> the min and max intensities are displayed
 * change the min and max intensities in 3D -> the 3D view is not updated, but the 2D ones are
 * switch to the 2D views, change the min/max intensities -> the 3D view is not updated, but the 2D ones are

This means we can only change the min and max intensities before switching for the first time in 3D.

Florent2305:
Fix #581 [Tests] 3.1.1 and master -> once min/max intensities in 3D, it's kept
For 4.0


ocommowi:
Easy, sounds good to me :) merging, thanks

PR #693, opened 2020-05-14
Put back algo name in undo redo toolbox

ocommowi:
Names of the algos run were not shown anymore in the undo redo toolbox after running a registration algorithm. This corrects it. To be pushed on 3.1.x as well


src/layers/legacy/medCoreLegacy/gui/toolboxes/medRegistrationSelectorToolBox.cpp
```diff
@@ -126,6 +126,9 @@ void medRegistrationSelectorToolBox::changeCurrentToolBox(int index)
     connect (currentToolBox(), SIGNAL (success()),this,SLOT(enableSelectorToolBox()));
     connect (currentToolBox(), SIGNAL (failure()),this,SLOT(enableSelectorToolBox()));
 
+    unsigned int index = this->comboBox()->currentIndex();
```
mathildemerle, line 129: index creates a compilation error since it's the name of a parameter of this function. I'm going to do PRs to solve that

Issue #695, opened 2020-05-14
DiffeomorphicDemons Freeze at the end


Florent2305:
some logs:

> DEBUG - Thu May 14 10:24:01 2020 - ""DiffeomorphicDemons\n  Max number of iterations   : [ 15, 10, 5 ]\n  Update rule   : DIFFEOMORPHIC\n  Gradient type   : SYMMETRIZED\n  Maximum step length   : 2 (voxel unit)\n  Update field standard deviation   : 0 (voxel unit)\n  Displacement field standard deviation   : 1.5 (voxel unit)\n  Use histogram matching   : false\n"" 
> DEBUG - Thu May 14 10:24:08 2020 - "Elasped time:  52.894" 
> DEBUG - Thu May 14 10:24:08 2020 - "medDatabaseNonPersistentImporter created with uuid: "{17ef33a1-04ae-4a4c-a089-72b95244e319}"" 
> INFO  - Thu May 14 10:24:08 2020 - "generating data for image factory" 
> DEBUG - Thu May 14 10:24:08 2020 - medAbstractImageData has no such property: "SeriesDescription"

data to reproduce bug:
https://filesender.renater.fr/?s=download&token=b6bcf1cb-0b5e-4be7-9395-2f6c48127b70

Florent2305:
At the end of diffeomorphic demons processus the call stack as follow:

|      FUNCTION                                                                               |   SOURCE FILE                                                                              | MODULE   |      DLL                       |
| ------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------ | -------- | ------------------------------ |
|  itk::ImageRegistrationFactory<itk::Image<float,3> >::GenerateData()                        | medinria_extprojs/src/rpi/registrationaddon/itkimageregistrationfactory.txx(45)            | RPI      | medRegistration.dll            |
|  itk::ProcessObject::UpdateOutputData(itk::DataObject * __formal)                           | medinria_extprojs/src/itk/modules/core/common/src/itkprocessobject.cxx(1778)               | ITK      | ITKCommon-5.0.dll              |
|  itk::DataObject::UpdateOutputData()                                                        | medinria_extprojs/src/itk/modules/core/common/src/itkdataobject.cxx(411)                   | ITK      | ITKCommon-5.0.dll              |
[...]

PR #703, opened 2020-05-25
medLoggerRedirectWithBoost

Florent2305:
This PR change the way to redirect cout and cerr.
Now don't use **dtk** to redirect log but **Boost**
It avoids some crashs or freezes when external dependencies, like RPI, did a cout or cerr.

The problem is due to incorrect re-lock mutex inside dtklogger
cd https://github.com/medInria/medInria-public/issues/695

### DIFF
--- a/src/app/medInria/medLogger.cpp
+++ b/src/app/medInria/medLogger.cpp
@@ -1,6 +1,10 @@
 #include "medLogger.h"
 
 #include <QTextStream>
+#include <iostream>
+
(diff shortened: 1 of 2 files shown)


Florent2305:
Just we will need to improve **line 24 of medLogger.cpp**
`medLoggerPrivate() : logFile(dtkLogPath(qApp).toLocal8Bit().data(), std::ios::app) {}`
to support the devilish utf-16 MS filesystem API

PR #602, opened 2020-04-13
prevent crash after loading : 1st nrrd image, 2: nhdr, 3 : nrrd, 

nbaraka:
I do not knwo why there this issue. 
Because if  you load first nhdr image then you do not this pb. 
I may hide a bug by preventing this crash. (smthing related LayerInfoVec in vtkImageView3D)

PR #706, opened 2020-05-26
Solve Glyph 2D into 3D rendering

Florent2305:
Solve: 
> 3D mode crash for Image with vector data or tensor when data is 2D: "itkDataImageVectorFloat3 #640


Florent2305:
We must retest #623 and #602

src/layers/legacy/medVtkInria/vtkImageView/vtkImageView3D.cxx
```diff
@@ -449,16 +449,10 @@ void vtkImageView3D::SetInput(vtkAlgorithmOutput* pi_poVtkAlgoOutput, vtkMatrix4
...
 
-            if (is3D())
-            {
-                this->InternalUpdate();
-            }
-            else
-            {
-                vtkErrorMacro( <<"Only 3D layers are supported in 3D orientation" );
-            }
+            special2DData();
```
mathildemerle, line 452: is it needed then?

ocommowi:
Sounds good to me data2DTreatment is not the nicest name but is3D was not either so good to go if it saves the day

Issue #696, opened 2020-05-14
Registration WS "adjustable wrench" icone don't respond without Image


Florent2305:
Into Registration workspace, the "adjustable wrench" icone has no action without image into the view.
It's a big problem because a user can want open an image via this icone.

mathildemerle:
In Filtering and Registration, the goal was to have fix containers, that's why the "Tools" button was disabled (but it would be better to display a gray icon then) as far as i remember. That's because we are not used to use the Open button in the Tools menu.

But then if you add a data, the right behavior appears: you can use the Tools button, but can't split or remove the views.

So, indeed there is a problem, and i would add 2 problems to solve. Summary of "Tools" menu issues:

 * In Filtering and Registration, we need to enable by default the Tool menu, but on "no split" mode, like when we add a data.
 * Add a right disabled Tools icon.
 * If you add a data in Registration for instance, you can Maximize it. If you remove it and add it back, you can't Maximize it. Choose if we want to be able to maximize or not in "no split" mode.

Issue #633, opened 2020-04-23
[FilteringWorkspaceL] remove data -> lose background 


mathildemerle:
In Filtering workspace, open a data, close it -> the background color is not the right one.

It happens on master and 3.1.1.

Issue #591, opened 2020-04-03
[Tests] 3.1.1 and master - last layer removed - display ghost button


mathildemerle:
In a workspace (Filtering, Meshing, etc):
 * open a data
 * remove it through the layer toolbox -> the "Choose a toolbox" widget ghost is displayed

This is not happening with the view close button. It does not happen all the time (1/3).
Could be a refresh/update problem.

mathildemerle:
In my humble opinion, we could put this issue on "Major" for the next release, after 3.1.1

mathildemerle:
solved

PR #712, opened 2020-05-29
[Vues] Tools button debugs - data/view signals during removal

mathildemerle:
Fix:
 * The `Close` view button is displayed disabled if the button is disabled (happens for instance on Maximize mode). Thus, I added a disabled close icon.
 * On views with `CLOSE_VIEW` mode, it is now possible to open a data through the `Tools` button, but you still can't split the views. Fix [Registration WS "adjustable wrench" icone don't respond without Image](https://github.com/medInria/medInria-public/issues/696).
 * As a consequence of the removal signals/functions cleaning, i separated the signals `contentChanged` and `viewRemoved`. Before that, in workspaces, each time you removed a view, you received also the `contentChanged` signal, which created some bugs and duplication of functions called.

Beginning:
 * I would have loved to solve completely [last layer removed - display ghost button](https://github.com/medInria/medInria-public/issues/591) but i managed at least to avoid problems in `CLOSE_CONTAINER` views, with removal through layers, and not the `Close` button. It's a complete maze. For instance `CLOSE_CONTAINER` or `CLOSE_VIEW` views have different removal behaviors, and removing data from the `Close` or `Layers` buttons are different also. And each behavior impact every issue listed in this PR.

Linked PR:
 * The changes in this PR shows a huge linked with [[FilteringWorkspaceL] remove data -> lose background](https://github.com/medInria/medInria-public/issues/633) also. I lost the new central widget for instance in `CLOSE_VIEW` views (aka "FIXED" text for instance in Registration workspace), because `d->view` at one point was not really null. I found a way to display it. However, it means that some problems are still here. Beware, every problems solved or not in this PR are linked :D.

PS: for now, the PR can't be tested because of a small QT problem on branch medInria3.1.1, but i'll rebase it when it's solved.

:m:


Florent2305:
Dans le workspace "Registration" 
- ajoute une image sur "fixed"
- supprimer l'image
- click sur l'onglet "Fuse"
-> crach ligne 89 de src/layers/legacy/medCoreLegacy/gui/toolboxes/medToolBoxBody.cpp

PR #719, opened 2020-06-03
Layers & Volume Rendering 

nbaraka:
**Improve layers management**: 

- layers are  independent (no re slice (it still could be done by a preprossessing  if needed), 

- the first image ROI do not have to include all the next one (feature asked by Im. Team) ; 

**Improve volume rendering** :   

- no limit on the number of volumes that could be included (used to be 4) 

- better opacity management

Please Tell if you have any regression in your workflow. 

Best regards.


src/layers/legacy/medVtkInria/vtkImageView/vtkImageView3D.cxx
```diff
@@ -41,564 +41,418 @@
...
+#include <vtkLookupTableManager.h>
+
+#include <vtkMatrix4x4.h>
 
 vtkStandardNewMacro(vtkImageView3D)
 
-//----------------------------------------------------------------------------
 vtkImageView3D::vtkImageView3D()
 {
-  this->VolumeProperty  = vtkVolumeProperty::New();
```
mathildemerle, line 50: is there changes in this file or is it only indentation changes?

src/layers/legacy/medVtkInria/vtkImageView/vtkImage3DDisplay.h
```diff
@@ -44,6 +45,12 @@ class vtkImage3DDisplay : public vtkObject
 
     vtkSetObjectMacro(LookupTable, vtkLookupTable);
     virtual vtkLookupTable*GetLookupTable();
+    void SetLookupTableName(QString name) {
```
mathildemerle, line 48: can you put the brackets under? Idem line 51

src/plugins/legacy/itkDataImage/interactors/medVtkViewItkDataImageInteractor.cpp
```diff
@@ -277,7 +277,7 @@ void medVtkViewItkDataImageInteractor::initParameters(medAbstractImageData* data
     }
     else
     {
-       this->opacityParameter()->setValue(1);
+       this->opacityParameter()->setValue(0.99);// if 1.0 : no signal is since the initial is 1.0 but it has been set without a emiting signal
```
mathildemerle, line 280: I do not understand i think. Are you using 0.99 to force the emit of the signal? Will it be set to 1 after?

mathildemerle:
Testing this PR, i have a crash opening a data, then its mesh, and then its mask. I tested on master branch, and i do not have this crash.

By the way, this PR would need to be switched to the master branch, not the 3.1.1 one. Maybe a rebase on master could solve the crash.

mjuhoor:
@nbaraka  would be interesting to update this PR. Would be cool to have this working.

PR #730, opened 2020-06-04
[FIX] - Add PixelDeviceRatio for OSX Screen Retina

juliencastelneau:
before using mapDisplayToWorldCoordinates  with MouseEvent
compute the device Pixel Ratio and apply to mouse Position


mathildemerle:
(NB, need to be done also on master)

PR #752, opened 2020-07-02
medStatusBar become threadsafe to prevent crash V3.1.x

### LINKED ISSUE #751: medStatusBar no thread-safe crash
medStatusBar contains 2 lists of messages (visible and hide).
If 2 threads or more change one of these list in same time, medInria can crash.

I detect this crash (it no just a code review) during importe many dicom series from a PACS and resize medInria window to a smallest size in the same time.
The signal shrinks the list of visible messages and grow-up the hide list, but the transfert from an to an other can conflict with life duration of a message.

**The crash is rare**

Florent2305:
fix #751

### DIFF
--- a/src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
+++ b/src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
@@ -22,7 +22,7 @@ public:
     QBoxLayout * statusBarLayout;
     int availableSpace; //available space on the status bar
     int statusBarWidth;
-    std::mutex listsofMsgMutex;      // mutex to guard conflict between external signals and/or internal update
+    std::recursive_mutex acccessMutex;      // mutex to guard conflict between external signals and/or internal update
     QList<medMessage *> hiddenMessageList; // storing messages waiting to be displayed
     QList<medMessage *> messageList;       // storing displayed messages
 };
@@ -32,7 +32,6 @@ medStatusBar::medStatusBar ( QWidget* parent ) : QStatusBar ( parent ), d ( new 
     d->statusBarLayout = NULL;
     d->statusBarWidth = this->size().width();
     d->availableSpace = -1; //value before initialization (after, always positive)
-    d->listsofMsgMutex;
 }
 
 medStatusBar::~medStatusBar()
@@ -69,13 +68,18 @@ QBoxLayout* medStatusBar::statusBarLayout()
  */
 void medStatusBar::setAvailableSpace ( int space )
 {
+    d->acccessMutex.lock();    
     d->statusBarWidth = this->size().width();
     d->availableSpace = space;
+    d->acccessMutex.unlock();
 }
 
 int medStatusBar::getAvailableSpace()
 {
-    return d->availableSpace;
+    d->acccessMutex.lock();
+    int iRes = d->availableSpace;
+    d->acccessMutex.unlock();
+    return iRes;
 }
 
 
@@ -84,7 +88,7 @@ int medStatusBar::getAvailableSpace()
 */
 void medStatusBar::init_availableSpace()
 {
-    if (d->availableSpace == -1 )
+    if (getAvailableSpace() == -1 ) // no need to lock mutex because is done by getAvailableSpace()
     {
         emit initializeAvailableSpace();
     } 
@@ -95,11 +99,13 @@ void medStatusBar::init_availableSpace()
 */
 void medStatusBar::spaceManagement()
 {
+    d->acccessMutex.lock();
     if (d->availableSpace != -1)       // if availableSpace initialized
     {
         d->availableSpace += (this->size().width()- d->statusBarWidth);    //update available space 
         d->statusBarWidth = this->size().width();                      // and statusbarWidth
-    } 
+    }
+    d->acccessMutex.unlock();
 }
 
 /**
@@ -109,13 +115,12 @@ void medStatusBar::spaceManagement()
 */
 void medStatusBar::addMessage ( medMessage* message )
 {
+    d->acccessMutex.lock();
     this->init_availableSpace();
-
     if ( message )
     {
         if ( d->statusBarLayout )
         {
-            d->listsofMsgMutex.lock();
             if ( d->availableSpace > message->size().width()+d->statusBarLayout->spacing())      // if enough space
             {
                 d->messageList.append(message);   // messages are stored in a list to easily hide newest messages 
@@ -127,10 +132,10 @@ void medStatusBar::addMessage ( medMessage* message )
             else
             {
                 d->hiddenMessageList.append(message);    //put the message into a list              
-            }
-            d->listsofMsgMutex.unlock();  
+            } 
         }
     }
+    d->acccessMutex.unlock(); 
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
```diff
@@ -14,18 +14,25 @@
...
 };
 
 medStatusBar::medStatusBar ( QWidget* parent ) : QStatusBar ( parent ), d ( new medStatusBarPrivate )
 {
     d->statusBarLayout = NULL;
-    this->statusBarWidth = this->size().width();
-    this->availableSpace = -1; //value before initialization (after, always positive)
+    d->statusBarWidth = this->size().width();
+    d->availableSpace = -1; //value before initialization (after, always positive)
+    d->listsofMsgMutex;
```
ocommowi, line 35: Heu ca sert a rien cette ligne la non ?

src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
```diff
@@ -147,34 +157,42 @@ void medStatusBar::removeMessage ( medMessage* message )
...
 */
 void medStatusBar::showHiddenMessage()
 {
-    if(!this->hiddenMessageList.isEmpty())                         // if message waiting to be displayed
+    d->listsofMsgMutex.lock();
+    if(!d->hiddenMessageList.isEmpty())                         // if message waiting to be displayed
     {
-        medMessage* messageToShow = hiddenMessageList.takeFirst();
+        medMessage* messageToShow = d->hiddenMessageList.takeFirst();
+        d->listsofMsgMutex.unlock();
```
ocommowi, line 181: Le unlock peut etre en dehors du if, tu le fais de toutes facons

src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
Florent2305: ha oui carrément c'était juste pour un point d’arrêt durant la mise au point, je vais le supprimer.

src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
Florent2305: Hélas, pas du tout, car l'appel à `addMessage(messageToShow);` ligne 168|182 lock le mutex et dans ce cas il y a deadlock.

src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
```diff
@@ -135,8 +143,10 @@ void medStatusBar::removeMessage ( medMessage* message )
 {
     if ( message )
     {
-        this->messageList.removeFirst();
-        this->availableSpace += (message->size().width()+d->statusBarLayout->spacing());         //update available space
+        d->listsofMsgMutex.lock();
+        d->messageList.removeFirst();
+        d->listsofMsgMutex.unlock();
```
mathildemerle, line 148: can you unlock after the space computation? As done in addMessage and hideMessage? In addMessage it's even done after the insertion of the widget

src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
ocommowi: Ah ok. Ben dans ce cas, lock le mutex dans le if comme ca tu peux virer le else

Florent2305: En fait si je l'ai fait avant le if c'est car dans le teste même du if on accède à la liste qui doit être protégée.
J'ai hélas vraiment pas eu  le choix de faire autrement.
Le Else je l'ai pas fait de gaieté de cœur hélas.

src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
fcollot: Mutexes must be unlocked as soon as possible to avoid unnecessary pauses in execution for the other threads waiting on this mutex.

src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
fcollot: Tu peux utiliser un `recursive_mutex` ce qui te permettrait sortir le unlock du if.

src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
Florent2305: @fcollot it's my point of view also
But @mathildemerle makes me notice that we can protect the `availableSpace `
I actually study if availableSpace concurrent access can create a graphical glitch...
@fcollot have you got an opinion about it ?

fcollot: In my first comment I was taking into account the fact that the goal of the mutex, as stated by its name, is to protect access to the message list only. It is true though that to avoid graphical glitches you could extend the protection and maybe in that case it would be simpler to do something like synchronized methods in java and systematically protect the whole body of the methods in this class (with recursive mutexes). There will no noticeable performance hit since messages will be sent at a much lower rate than what the multithreaded execution can handle.

fcollot: (so my first message was more of a general best practices comment, not specifically related to the fact there is no real critical performance issue here)

Florent2305: Of course, I will evolve the mutex name if the goal of mutex extends.
I'm totally agree with the best practice you highlight.
I think your approche is good.
Normally I refrain from using recursive mutex. Few perform and often symptomatic of poor design. As explained in the post https://stackoverflow.com/questions/14498892/stdmutex-vs-stdrecursive-mutex-as-class-member/28741305
But there, it is true that in any case the disgne was not thought for , and that the problems of preformance are so far ...
So I think I'm going to make an exception to my rule;)

fcollot: Thank you for the link. I agree that you can avoid recursive mutexes by redesigning the code. On the other hand, I have noticed that most of the questions asked on stack overflow (or similar forums) are answered by something like "why do you want to do that, it is bad design". This is especially true for Python questions 😄.

src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
mathildemerle: You forgot to remove this line ^^

src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
mathildemerle: the unlock() still could be moved after the next line to avoid bugs with the available space computation

src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
Florent2305: done

src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
Florent2305: done

src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
Florent2305: done

mathildemerle (approved):

PR #753, opened 2020-07-02
medStatusBar become threadsafe to prevent crash V4

### LINKED ISSUE #751: medStatusBar no thread-safe crash

Florent2305:
Fix #751

### DIFF
--- a/src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
+++ b/src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
@@ -14,18 +14,25 @@
 #include <medStatusBar.h>
 
 #include <QtCore>
+#include <mutex>
 
 class medStatusBarPrivate
 {
 public:
     QBoxLayout * statusBarLayout;
+    int availableSpace; //available space on the status bar
+    int statusBarWidth;
+    std::mutex listsofMsgMutex;      // mutex to guard conflict between external signals and/or internal update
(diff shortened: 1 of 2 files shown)


src/layers/legacy/medCoreLegacy/gui/commonWidgets/medStatusBar.cpp
```diff
@@ -14,18 +14,24 @@
...
+    QList<medMessage *> hiddenMessageList; // storing messages waiting to be displayed
+    QList<medMessage *> messageList;       // storing displayed messages
 };
 
 medStatusBar::medStatusBar ( QWidget* parent ) : QStatusBar ( parent ), d ( new medStatusBarPrivate )
 {
-    d->statusBarLayout = nullptr;
-    this->statusBarWidth = this->size().width();
-    this->availableSpace = -1; //value before initialization (after, always positive)
+    d->statusBarLayout = NULL;
```
mathildemerle, line 32: beware of the NULL regression

Florent2305: done

mathildemerle (approved):

PR #754, opened 2020-07-08
[VTK9.0.1] base to switch medInria on it

mathildemerle:
Need https://github.com/Inria-Asclepios/TTK-Public/pull/13

I was working on something else, and i wanted to test VTK 9.0.1 with it. At the end, i'm on holidays on friday evening until 14/08, i do not have time to finished right now, and i think i do not need it for my work, but the switch to 9.0.1 is started, and it could be interesting to keep the code as a base.

If you want to debug, try the VTK functionalities, on different OS, you can update this PR, no problem, or create new PRs if needed. I'm putting it at "Invalid" for now. It compiles on my side (minus i need to check manually VTK_GROUP_ENABLE_Qt in the cmake of VTK for now), but i did not tested.

We can discuss the changes.

:m:


src/layers/legacy/medLog/CMakeLists.txt
```diff
@@ -28,9 +28,8 @@ find_package(ITK REQUIRED COMPONENTS ITKCommon)
 include(${ITK_USE_FILE})
 
 # find vtk
-find_package(VTK REQUIRED COMPONENTS vtkCommonCore)
-include(${VTK_USE_FILE})
-
+find_package(VTK REQUIRED COMPONENTS CommonCore)
+include_directories(${VTK_INCLUDE_DIRS})
```
ocommowi, line 32: This line (32) shouldn't be required (applies to other files as well). Normally the fact that you include a VTK module in your target links is enough. At least it worked for me when translating anima to VTK 9

ocommowi (commented):
Just a quick comment on the include of VTK_INCLUDE_DIRS which I think are not necessary. Would be worth trying without them, it worked for me without them on another repo (anima)

src/layers/legacy/medLog/CMakeLists.txt
mathildemerle: Thanks for the tip, i'll test that :)

PR #755, opened 2020-07-10
[VideoExport] add a video export feature: avi, ogv, jpeg

mathildemerle:
This PR adds:
 * Export of **ogv** or ~~**avi**~~ **mp4** video of a view through **FFMPEG** (Linux and Mac) ~~or **MPEG2** (Windows)~~.
 * Export of a batch of **jpeg** of a view.
 * Kind of export: **rotation** of a mesh.
 * Kind of export: **time** scrolling for a 4D data, with display of the time on the view.
 * Kind of export: **slice** scrolling for a data (for _MUSIC_ devs: it's a new feature, and by the way i switched also ffmpeg from 0.7 to 4.3).
 * The video button is next to the screenshot button, in the bottom bar of the application.

FFMPEG or MPEG2 lib are compiled according to your OS, and are linked to VTK. So, you'll need to recompile from scratch at least VTK in order to test this.

It would be good to test this PR at least on Mac and Windows, since I work on Linux. In MUSIC we did not use fibers/vectors data so, it could be interesting to test some fibers/vector data too.

We had problems with FFMPEG 0.7 with VTK ~5 on Windows, that's why we were using MPEG2 only on Windows. We had also problems with MPEG2 on Linux or Mac haha. That was complicated. It could be solved in VTK 8 or 9, perhaps, I don't know.

Later, it could be interesting to enhance the display of the computation, on Ubuntu 20.04 my cursor does not change when a computation is currently running, only the logs are showing that something is happening. It's a common problem in the application, that's why we added by the way the deactivation of toolboxes during computation. Maybe a message displayed in the current log bar? We'll see that.

:m:


Florent2305:
I am embarrassed.

MPEG2 library readme indicates patent and licensing issues.
We also have no redistribution rights (compiled or not) provided by disclaimers of headers of the .h files.

Moreover, ffmpeg depending on its compilation flag can switch from LGPL to GPL, which is not possible. I will investigate this question.

Sincerely sorry. Hope I would find a workaround like only use LGPL part of ffmpeg in any cases.

mathildemerle:
Update: MPEG2 Writer has been removed from VTK >= 8.2. We were using it on Windows, so, it has been removed too.

src/layers/legacy/medCoreLegacy/CMakeLists.txt
```diff
@@ -25,8 +25,6 @@ find_package(OpenGL REQUIRED)
 
 list_source_files(${TARGET_NAME}
   ${CMAKE_CURRENT_SOURCE_DIR}
-  threadpool/threadpoolLight
```
Florent2305: Oui

mathildemerle:
I added a commit which:
 * solves a problem with the "if ffmpeg", 
 * adds waiting cursors, 
 * solves a number of step initialize error, 
 * adds a quality parameter to ffmpeg, 
 * switches .avi to .mp4, for ffmpeg,
 * and adds an explanation for users telling them that they can use their own tools for the jpeg or the videos.

src/plugins/legacy/medExportVideo/medExportVideo.cpp
```diff
@@ -0,0 +1,425 @@
...
+    source->SetNumberOfScalarComponents(3);
+    source->SetExtent(0, d->width-1, 0, d->height-1, 0, 0);
+
+    vtkSmartPointer<vtkGenericMovieWriter> writerVideo;
+
+    if (d->format == OGGVORBIS)
+    {
+        vtkSmartPointer<vtkOggTheoraWriter> writerVideoTmp = vtkSmartPointer<vtkOggTheoraWriter>::New();
+        writerVideoTmp->SetInputConnection(source->GetOutputPort());
+        writerVideoTmp->SetFileName(d->filename.toStdString().c_str());
```
Florent2305, line 230: Be careful with path encoding

src/plugins/legacy/medExportVideo/medExportVideo.cpp
```diff
@@ -0,0 +1,425 @@
...
+        writerVideoTmp->SetRate(d->frameRate);
+        writerVideoTmp->SetSubsampling(d->subsampling);
+        writerVideoTmp->SetQuality(d->quality);
+
+        writerVideo = writerVideoTmp;
+    }
+    else
+    {
```
Florent2305, line 238: Be careful!
What appends if entre in with:

- d->format with other value than FFMPEG
- d->format at FFMPEG but with MED_USE_FFmpeg not define

suggestion:
 
`#ifdef (MED_USE_FFmpeg)`
`        else if (d->format == FFMPEG)`
`        {`
`             vtkSmartPointer<vtkFFMPEGWriter> writerVideoTmp = vtkSmartPointer<vtkFFMPEGWriter>::New();`
`             writerVideoTmp->SetInputConnection(source->GetOutputPort());`
`             writerVideoTmp->SetFileName(d->filename.toStdString().c_str());`
`             writerVideoTmp->SetRate(d->frameRate);`
`             writerVideoTmp->SetQuality(d->quality);`
`             writerVideo = writerVideoTmp;`
`        }`
`#endif`
`        else`
`        {`
`            return medAbstractProcessLegacy::FAILURE;`
`        }`
`    }`

src/plugins/legacy/medExportVideo/medExportVideo.cpp
```diff
@@ -0,0 +1,425 @@
...
+#if defined (MED_USE_FFmpeg)
+        if (d->format == FFMPEG)
+        {
+            vtkSmartPointer<vtkFFMPEGWriter> writerVideoTmp = vtkSmartPointer<vtkFFMPEGWriter>::New();
+            writerVideoTmp->SetInputConnection(source->GetOutputPort());
+            writerVideoTmp->SetFileName(d->filename.toStdString().c_str());
```
Florent2305, line 244: Be careful with path encoding

src/plugins/legacy/medExportVideo/medExportVideo.cpp
```diff
@@ -0,0 +1,425 @@
...
+    d->formatComboBox = new QComboBox(d->exportDialog);
+    // Free, opensource, no patent.
+    // "Before 2007, the .ogg filename extension was used for all files whose content used the Ogg container
+    // format. Since 2007, the Xiph.Org Foundation recommends that .ogg only be used for Ogg Vorbis audio files.
+    // The Xiph.Org Foundation decided to create a new set of file extensions and media types to describe different
+    // types of content such as .oga for audio only files, .ogv for video with or without sound (including Theora),
+    // and .ogx for multiplexed Ogg."
+    d->formatComboBox->addItem("Ogg Vorbis (.ogv)", OGGVORBIS);
+    d->formatComboBox->addItem("JPEG (.jpg .jpeg)", JPGBATCH);
+#if defined (MED_USE_FFmpeg)
```
Florent2305, line 303: same as previous

src/plugins/legacy/medExportVideo/medExportVideo.h
```diff
@@ -0,0 +1,62 @@
...
+    virtual QString description() const;
+    virtual QString identifier() const;
+    static bool registered();
+
+    enum VideoFormat
+    {
+        OGGVORBIS,
+        JPGBATCH,
+        FFMPEG,
+        MPEG2
```
Florent2305, line 29: remove MPEG2

superbuild/projects_modules/VTK.cmake
```diff
@@ -49,7 +49,15 @@ set(git_tag v8.1.2)
...
   set(${ep}_cxx_flags "${${ep}_cxx_flags} -w")
-  # set(unix_additional_args -DVTK_USE_NVCONTROL:BOOL=ON)
+endif()
+
+# library extension
+if (UNIX AND NOT APPLE)
+    set(extention so)
+elseif(APPLE)
+    set(extention dylib)
+elseif (WIN32)
```
Florent2305, line 59: *you can remove WIN32 Close

Florent2305 (changes requested):
I made few comments and remarks.
I will test it on Windows

Florent2305 (approved):
Nikel

Florent2305 (approved):
ok

Florent2305 (approved):
Ok but not really useful to change DCMTK.cmake and ITK.cmake

superbuild/patches/VTK.patch
```diff
@@ -0,0 +1,34 @@
...
++
+ #include <ctime>
+ 
+ //---------------------------------------------------------------------------
+@@ -190,7 +192,7 @@ int vtkOggTheoraWriterInternal::Start()
+   th_info_clear(&thInfo);
+ 
+   // Finally, open the file and start it off.
+-  this->outFile = fopen(this->Writer->GetFileName(),"wb");
++  this->outFile = vtksys::SystemTools::Fopen(this->Writer->GetFileName(),"wb");
```
mathildemerle, line 28: explanation: we found with @Florent2305 a bug in VTK in vtkOggTheoraWriter on Windows (classic UTF16 problem). There is no problem on vtkJPEGWriter, which uses vtksys::SystemTools::Fopen already. This bug is classic also in other parts of VTK

Issue #460, opened 2019-10-24
[Generic Workspace] use as medSelectorWorkspaces


mathildemerle:
A discussion about the current and/or future workspaces.

"Legacy" selector workspaces can be generated and registered in plugins (plugins here, or in external projects). For instance, Filtering, Meshing, Reformat, Registration and Segmentation are plugins on this repository.

The Generic Workspace (which is "non-Legacy", or "dtk2") buttons in the Settings buttons on homepage under medInria description could perhaps in the future be removed to let users create their own workspaces as they want.

Florent2305:
The generic Workspace  like it is actually will be remove on 4.1

But keep the concept of generic Workspace can be a good way. This to create a plugin without re-generate a complet Workspace but just custom a "new generic Workspace".

Unless the gain is too small ...

PR #759, opened 2020-08-20
[Generic Workspace] use medSelectorWorkspace instead

mathildemerle:
As suggested in https://github.com/medInria/medInria-public/issues/460 this PR removes the Generic Workspace tool which is redundant with medSelectorWorkspace workspaces that we can create as we want. Moreover, we added a lot of dedicated workspaces since some releases of medInria, which allow the creators of plugins to choose an already existent workspace for their plugin.

:m:


Florent2305 (approved):
ok

PR #761, opened 2020-08-24
Add dev rules on branches to CONTRIBUTING.md

dmargery:
Attempt to make branching and development rules for different branches as explicit as possible

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,16 @@
 
 Welcome to the medInria Community! 
 
+# Getting started
+
+Please **read carefully** [the GitHub guide on Contributing to Open Source](https://guides.github.com/activities/contributing-to-open-source/). We also recommend you to check this [more detailed documentation on issues and pull requests](https://help.github.com/categories/collaborating-with-issues-and-pull-requests/).
+
+# How development is organized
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -2,6 +2,16 @@
...
+
+# How development is organized
+- The `master` branch hosts the code under development. We aim to only merge stable contributions, but only minimal effort is put to check side effects of these contributions (ie. contributed fixes or features are tested before a merge, but we acknowledge they might breack things).
+- If you build application based on medInria, we recommend you base your code on medInria's stable branches (those named after the version they prepare as in `3.1.x`), not on `master`, as these see all the systematic testing.
+- Renaming, reorganizing and similar janitoring on the code will only be accepted during short periods before a new stable branch is created. This helps the flow of code between the stable branch and the `master` branch when fixes are applied to one of them.
+- When cherry-picking between branches fails, and you port a commit from one branch to the otherdirectly, the commit message shoud reference the commit that has been ported with the same message with `ported commit <git hash>` appended.
```
mathildemerle, line 13: Do you mean:

> - Sometimes cherry-picking between branches fails. In that case, the port of a Pull Request (one commit or more) from one branch to an other has to be done manually. You should add the original Pull Request link in the new Pull Request message.

? Because we don't send commits directly on the branches, we use Pull Requests.

PR #765, opened 2020-09-02
[vtkDataMeshInteractor] Fix mesh LUT messing with image LUT

mjuhoor:
- Fixes a bug where, if there is a volume and a mesh in the 2D view, changing the LUT of the mesh would also change the LUT of the image.


Florent2305 (approved):
Ok

PR #767, opened 2020-10-02
[MeshManipulation] Transform properly applied to EP maps

mjuhoor:
- Mesh manipulation applies transformations to EP maps


mathildemerle (approved):

paulineMig (approved):

Florent2305 (approved):
ok

PR #774, opened 2020-10-14
[ExternalProjects] avoid duplication with dtkImaging

mathildemerle:
Avoid a duplication of external projects lines with or without dtkImaging.

:m:


Florent2305 (approved):
ok

PR #778, opened 2020-10-28
[FFMPEG] switch to version n3.4.8

mathildemerle:
After some full continuous integration compilations, switch ffmpeg to the n3.4.8 version (which, by the way, is the one used in apt versions in Ubuntu20).

:m:


Florent2305 (approved):
Ok

Issue #790, opened 2020-11-10
[Paint] after first drawing click, crash continuing drawing


mathildemerle:
[Test of the Ubuntu20 Jenkins binary on a Ubuntu20 computer]

 * Open a data in Paint Segmentation
 * Change the contrast ? Do quickly the segmentation ?
 * Click on the image: a circle is drawn
 * Randomly, you can't continue the drawing, and the application crashes:

```
med::AlgorithmPaintToolBox::isMask2dOnSlice() () from /home/mathildemerle/Dev/tmp/MUSICardio-3.0.0-Ubuntu_20.04-x86_64/bin/plugins_legacy/libmedAlgorithmPaint.so
med::AlgorithmPaintToolBox::updateStroke(med::ClickAndMoveEventFilter*, medAbstractImageView*)
med::ClickAndMoveEventFilter::mouseMoveEvent(medAbstractView*, QMouseEvent*)
```

PS: same crash on medInria and MUSICardio

By the way, the drawing creates a warning "Paint Segmentation: This toolbox is designed to be used with 3D volumes" because of the ROI type.

PR #792, opened 2020-11-10
[PaintToolBox]: fix crash

paulineMig:
fix https://github.com/medInria/medInria-public/issues/790

### DIFF
--- a/src/plugins/legacy/medAlgorithmPaint/medAlgorithmPaintToolBox.cpp
+++ b/src/plugins/legacy/medAlgorithmPaint/medAlgorithmPaintToolBox.cpp
@@ -1262,7 +1262,7 @@ AlgorithmPaintToolBox::GenerateMinMaxValuesFromImage ()
 
 void AlgorithmPaintToolBox::updateStroke(ClickAndMoveEventFilter * filter, medAbstractImageView * view)
 {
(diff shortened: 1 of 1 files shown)


mathildemerle (approved):

mjuhoor (approved):

nebatmusic (changes requested):
I think the pb is in the caller of this function. 

This is strange ? 
void AlgorithmPaintToolBox::updateStroke(ClickAndMoveEventFilter * filter, medAbstractImageView * view)
{
    setCurrentView(currentView);
....

paulineMig:
I agree this is really strange (because it put currentView to currentView) but I still had the bug with setCurrentView(view). I will do some more testing and check everything else if nothing is broken i will add the modification.

paulineMig:
> I think there is also a pb is in the caller of this function.
> 
> This is strange ?
> void AlgorithmPaintToolBox::updateStroke(ClickAndMoveEventFilter * filter, medAbstractImageView * view)
> {
> setCurrentView(currentView);
> if ( !isMask2dOnSlice() )
> {
> return;
> }
> ....
> }
> It seems that is should be setCurrentView(view);

Done

mathildemerle (approved):

nebatmusic (approved):

PR #800, opened 2020-11-12
improve cmake file on legacy code

nebatmusic:
remove un necessary include directories. 
Since during a target_link on a modern cmake project (like medInria) you have automatically what you need to build. (headers, flags that have been made PUBLIC). 
I didn't build this one (I only retrieve most of the changes done on https://github.com/Inria-Asclepios/medInria-public/pull/665)

### DIFF
--- a/src/layers/legacy/medImageIO/CMakeLists.txt
+++ b/src/layers/legacy/medImageIO/CMakeLists.txt
@@ -12,18 +12,27 @@
 # PURPOSE.
 #
 ################################################################################
-
 set(TARGET_NAME medImageIO)
-
 
 ## #############################################################################
 ## find requireds
 ## #############################################################################
+set(ITK_IMAGES_DEPENDENCIES
+    ITKIOImageBase
+    ITKIOMeta
+    #common formats for common images
+    ITKIOBMP ITKIOJPEG ITKIOJPEG2000 ITKIOPNG ITKIOTIFF
+    #Medical images formats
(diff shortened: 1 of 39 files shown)


mathildemerle:
I'm sorry but we are in a pre-release phase, doing tests and debugging. We try really hard to compile our app on Linux/Mac and Windows since weeks, and we managed to do a lot to succeed. I agree with @Florent2305 in https://github.com/Inria-Asclepios/medInria-public/pull/665 saying that it's a huge change and not a debug, and it would be better in 4.0.

nebatmusic:
_> I'm sorry but we are in a pre-release phase, doing tests and debugging. **We try really hard to compile our app on Linux/Mac and Windows since weeks, and we managed to do a lot to succeed**._ 
The purpose of this change is to improve our build files in order to improve the way we 're using the build generation tool. The job that has been done in most part of medinria public.  (remove old cmake syntax to the new cmake paradigm). 

_I agree with @Florent2305 in [Inria-Asclepios#665](https://github.com/Inria-Asclepios/medInria-public/pull/665) saying that it's a huge change and not a debug, and it would be better in 4.0._

Could you tell what makes you think that is a huge change ? 
the number of files, the number of lines or the time at which you read this PR, your mood ....

juliencastelneau:
I think Florent (and Mathilde) are agree with your idea.
They just say that it is better to introduce it in medInria v4.0

medInria3.2 is a transitional version of medInria.
The main goal is to provide a usable software for clinicians and researchers as soon as possible.
The enhancements related to modernization of our build tools might be done in 4.0.

Florent2305:
Keep idea for V4
Create equivalent PR on V4

PR #803, opened 2020-11-20
[StatROI] variance was a std deviation computation

mathildemerle:
statsROI::runMeanVariance function was badly named and computes mean and standard deviation.

:m:


mjuhoor (approved):

Florent2305 (approved):
ok

PR #807, opened 2020-11-26
[DICOM import] Quick fix for creation of multi series from unique series

### LINKED ISSUE #806: [DICOM] - Import 2 series while there is only one
- Go to Browser
- Choose a dicom directory (with 16 dicom file for instance)
- Import (or temporary import)
- open Visualisation workspace
==> In Patient/Study, you can see 2 series with 8 slices each.

It is not the case with all data.
I experienced it with a set of data. Try to import in Slicer or read with a python script. There is only one serie...
[100.000000-DCE Subtraction-32256.zip](https://github.com/medInria/medInria-public/files/5597859/100.000000-DCE.Subtraction-32256.zip)

mjuhoor:
- Fixes #806

**This is a quick fix that does not adress the underlying problem.**

Simplified description of the issue:
When looking for files to import, medInria creates a temporary 'unique' identifier for each image series: if `n` files have the same identifier then they are part of the same series.
This identifier is a concatenation of multiple values of metadata, like patient name, series description, slice thickness, rows, columns... and orientation
The problem is that floating point values (like orientation) can have small variations in each DICOM files: this creates different identifiers for the same image series.
The issue was thought 'solved' previously, as evidenced by the comments in the code. 

I personally believe that we should be creating an identifier using other metadata, e.g. the [Series Instance UID](https://dicom.innolitics.com/ciods/ct-image/general-series/0020000e) tag and others, and remove the decimal tags.

### DIFF
--- a/src/layers/legacy/medCoreLegacy/database/medAbstractDatabaseImporter.cpp
+++ b/src/layers/legacy/medCoreLegacy/database/medAbstractDatabaseImporter.cpp
@@ -251,6 +251,7 @@ void medAbstractDatabaseImporter::importFile ( void )
             if ( !volumeUniqueIdToVolumeNumber.contains ( volumeId ) )
             {
                 volumeUniqueIdToVolumeNumber[volumeId] = volumeNumber;
+                std::cout<<" -------- adding "<<volumeId.toStdString()<<" with "<<volumeNumber<<std::endl;
                 volumeNumber++;
             }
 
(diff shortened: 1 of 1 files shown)


mathildemerle (approved):

Florent2305 (approved):
Ok

Florent2305:
> I personally believe that we should be creating an identifier using other metadata, e.g. the [Series Instance UID](https://dicom.innolitics.com/ciods/ct-image/general-series/0020000e) tag and others, and remove the decimal tags.

I'm agree with this do you think it's a big change ?

msermesant:
be careful here. some of this code was specifically designed to reconstruct 3D+ t images from multiple 2D+t series, this is why it is also looking at position and orientation. This is typically the case with cine MRI. Please ensure that we still have a way to do this (it could be done in a 2nd step after importing all the series).

Florent2305:
> be careful here. some of this code was specifically designed to reconstruct 3D+ t images from multiple 2D+t series, this is why it is also looking at position and orientation. This is typically the case with cine MRI. Please ensure that we still have a way to do this (it could be done in a 2nd step after importing all the series).

I can discuss with Olivier to find the best way to fix this bug while preserving this feature.

PR #808, opened 2020-11-27
Fix Externals projects (medInria itself) forgot paths to dependencies.

Florent2305:
Eg:
Build whole the super-project.
Now, try to configure medInria itself, with CMake-gui for example, to change something.
In this case, the first time, medInria forgot paths to ITK, VTK, ...

This because during the super-build, medInria already has paths from the Super-build to external dependencies.
So, it didn't perform entirely find_package(s), and the consequence is paths are not save into local cache


juliencastelneau (approved):

superbuild/projects_modules/ITK.cmake
```diff
@@ -68,6 +68,10 @@ set(cmake_args
   -DITK_LEGACY_REMOVE:BOOL=ON
   -DVTK_DIR:PATH=${VTK_DIR}
```
ocommowi, line 69: Can be removed from cmake_args

ocommowi (commented):
Works for me, just one comment to resolve. Just a silly question: why not using cmake_cache_args for everything ?

superbuild/projects_modules/ITK.cmake
Florent2305: Good question :)
4 cases:
- Other variables must be defined via cmake_cache_arg and therefore, indeed, we must change (I did not detect this type of variables)
- Some variables are marked "cache" in the destination cmake. In this case, cmake_cache_arg and cmake_arg provide the same behavior.
- Maybe for some variables, they should not be cached (I don't know which ones might be affected)
- Variables need to be in "cache" and are already into the cmake_cache_arg by this PR.

For the moment, this modification is enough. If you think about variables cases which need to be put into cmake_cache_arg, let me know.
If we have only case 2 and 4 we can juste use cmake_cache_arg.
If we have case 1 we must determine, before all, why we didn't already detect bugs before.

ocommowi: Well from what I see in the use of cmake_args right now, all variables defined are in fact kept in cache since they are part of the destination variables that are defined anyway. So we could put everything as cmake_cache_args in my opinion. But we can also keep it as is for now and see this later when we modernize cmake files.

If you can just remove line 69 of the ITK.cmake file (redundant with later definition line 73 in cmake_cache_args), then the PR is acceptable to me.

Florent2305: Ho yes I didn't see the code duplicate.
Thank you

Florent2305: I will put everything as cmake_cache_args in a second phase according with you.

Florent2305: here the correction commit:
https://github.com/medInria/medInria-public/pull/808/commits/1854f9c60b29bb2ea3193c90c32b0495a708e23e

ocommowi (approved):

PR #809, opened 2020-11-30
[Database importer] Fix creation of multiple clones of the same study when importing.

### LINKED ISSUE #757: [Database] Saving to permanent database fails to find existing patient or study
Steps:
- Assume the database has already the patient named "John Doe", with an existing study "Empty study"
- Temporary import any 2 meshes : these are put in the temporary  _John Doe_ patient, _Empty study_ study.
- Save the first mesh: it is not saved in the existing "Empty Study". Instead it creates a duplicate one.
- Save the second mesh: again, a third "Empty Study" is created.

Additional info:
Sometimes even the patient is duplicated.

Fails for images and other types too, not just meshes
Don't know if it's a real database bug, or just a 'visual' bug related to the QTreeWidget items.
Tested with current medInria 3.1.1 release, and MUSIC current dev. branch.

mjuhoor:
- Fixes #757 , so it seems

Instanciating a QString with no parameters and "" (empty string) is not the same thing. This seems important when querying the database for existing studies, patients, etc.

Does this change impacts other areas of the code ?

### DIFF
--- a/src/layers/legacy/medCoreLegacy/data/medMetaDataKeys.h
+++ b/src/layers/legacy/medCoreLegacy/data/medMetaDataKeys.h
@@ -48,7 +48,7 @@ namespace medMetaDataKeys
 
         const QStringList getValues(const medAbstractData *data) const { return data->metaDataValues(KEY); }
 
-        const QString getFirstValue(const medAbstractData *data, const QString defaultValue=QString()) const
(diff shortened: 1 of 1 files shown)


mathildemerle:
Indeed, QString() seems to be initialized to null, not an empty string

mathildemerle (approved):

Florent2305 (approved):
ok

Issue #791, opened 2020-11-10
[PolygonROI] ROI -> 3D -> close ROI -> crash


mathildemerle:
[Test of the Ubuntu20 Jenkins binary on a Ubuntu20 computer]

 * Open a data in PolygonRoi
 * Activate the toolbox
 * Do a ROI
 * Click on the 3D button inside the toolbox (which has a graphical glitch by the way)
 * Close the view with the ROI
 * Enjoy a good crash:

```
medViewParameterGroupL::removeImpactedView(medAbstractView*) ()
   from /home/mathildemerle/Dev/tmp/MUSICardio-3.0.0-Ubuntu_20.04-x86_64/lib/libmedCoreLegacy.so
medViewParameterGroupL::~medViewParameterGroupL() ()
```
Warning in the log: `"polygonRoiToolBox :: clickClosePolygon - no view in container."`

Same crash on medInria and MUSICardio.

Issue #720, opened 2020-06-03
[PolygonROI] 3D icon cut


mathildemerle:
This is a really small problem, but i've got:

![polygonRoi-3DiconCut](https://user-images.githubusercontent.com/3910352/83611273-07e47b80-a581-11ea-8a90-43c363d07a03.png)

The 3D icon is cut.

PR #817, opened 2020-12-08
[PolygonROI] solve graphical/ergonomy pb and crashes (split, clear)

mathildemerle:
Fix https://github.com/medInria/medInria-public/issues/791
Fix https://github.com/medInria/medInria-public/issues/720

I wanted to solve the 2 bugs from the issue 791 in PolygonROI:
 * Open a data, split the view with the orientation buttons, close the first view -> crash
 * The orientation buttons have a glitch, they are cut (on my side at least)

But i found some other problems, which **are** solved also on this PR:
 * Open a data, activate PolygonROI, draw a ROI, change the current toolbox -> the ROI and tick are not cleared
 * Open a data, activate, split, change the current toolbox, go back to PolygonROI -> the toolbox is blocked on deactivate mode
 * Open a data, activate, split, close the application -> crash
 * Solve some typos and ergonomic position of some buttons to a better understanding of the use of this toolbox
 * PolygonROI -> Activate -> Split -> switch to an other tlbx -> go back to PolygonROI -> can't use the tlbx until the user remove the 2nd view
 * PolygonROI -> Activate -> Split -> remove the 2nd view -> remove the first one -> the remaining view is corrupted

I found also some other problems that are **not** solved by this PR:
 * Open a data, activate, split, remove the layer by the cross in the Layer toolbox -> crash
 * Open VOI Cutter firstly -> add data in the view -> switch to PolygonROI -> split -> crash
 
Hope this PR will help, do not hesitate to test or comment.

Capture d’écran du 2020-12-08 15-50-12

:m:

### DIFF
--- a/src/plugins/legacy/polygonRoi/toolboxes/polygonRoiToolBox.cpp
+++ b/src/plugins/legacy/polygonRoi/toolboxes/polygonRoiToolBox.cpp
@@ -445,14 +445,20 @@ void polygonRoiToolBox::updateTableWidgetView(unsigned int row, unsigned int col
 
         // Closing behavior of the non-main view
         medAbstractImageView* view = static_cast<medAbstractImageView *> (container->view());
-        connect(view, &medAbstractImageView::closed, [this, mainContainer]()
+        connect(view, &medAbstractImageView::closed, this, [this]()
         {
-            if (viewEventFilter && mainContainer)
+            if (viewEventFilter)
             {
                 viewEventFilter->clearAlternativeView();
-                mainContainer->setSelected(true);
-                // Once the alternate container is closed, the main one is allowed to be closed
-                mainContainer->setClosingMode(medViewContainer::CLOSE_VIEW);
+                medTabbedViewContainers *tabs = this->getWorkspace()->tabbedViewContainers();
+                if (tabs)
+                {
+                    QList<medViewContainer*> containersInTab = tabs->containersInTab(tabs->currentIndex());
+
+                    containersInTab.at(0)->setSelected(true);
+                    // Once the alternate container is closed, the main one is allowed to be closed
+                    containersInTab.at(0)->setClosingMode(medViewContainer::CLOSE_VIEW);
+                }
             }
         });
 
--- a/src/plugins/legacy/polygonRoi/toolboxes/polygonRoiToolBox.cpp
+++ b/src/plugins/legacy/polygonRoi/toolboxes/polygonRoiToolBox.cpp
@@ -45,7 +45,7 @@ polygonRoiToolBox::polygonRoiToolBox(QWidget *parent ) :
     displayWidget->setLayout(layout);
 
     activateTBButton = new QPushButton(tr("Activate Toolbox"));
-    activateTBButton->setToolTip(tr("Activate closed polygon mode"));
+    activateTBButton->setToolTip(tr("Activate closed polygon mode. You should only have one view."));
     activateTBButton->setCheckable(true);
     activateTBButton->setObjectName("closedPolygonButton");
     connect(activateTBButton,SIGNAL(toggled(bool)),this,SLOT(clickClosePolygon(bool)));
@@ -194,6 +194,10 @@ void polygonRoiToolBox::updateView()
             return;
         }
         medAbstractImageView *v = qobject_cast<medAbstractImageView*>(view);
+        if (!v)
+        {
+            return;
+        }
 
         if (v->layersCount()==1 && viewEventFilter)
         {
@@ -235,6 +239,7 @@ void polygonRoiToolBox::updateView()
                 emit currentLabelsDisplayed();
 
                 updateTableWidgetItems();
+
                 connect(currentView, SIGNAL(layerRemoved(uint)), this, SLOT(onLayerClosed(uint)), Qt::UniqueConnection);
                 connect(view, SIGNAL(orientationChanged()), this, SLOT(updateTableWidgetItems()), Qt::UniqueConnection);
                 connect(view, SIGNAL(orientationChanged()), this, SLOT(manageTick()), Qt::UniqueConnection);
@@ -255,7 +260,7 @@ void polygonRoiToolBox::onLayerClosed(uint index)
             viewEventFilter->removeFromAllViews();
             viewEventFilter->reset();
         }
-        clear();
+
         medTabbedViewContainers *tabs = this->getWorkspace()->tabbedViewContainers();
         if (tabs)
         {
@@ -271,6 +276,8 @@ void polygonRoiToolBox::onLayerClosed(uint index)
                 }
             }
         }
+
+        clear();
     }
 }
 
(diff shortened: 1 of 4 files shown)


paulineMig (approved):

src/plugins/legacy/polygonRoi/toolboxes/polygonRoiToolBox.cpp
```diff
@@ -135,6 +137,7 @@ polygonRoiToolBox::polygonRoiToolBox(QWidget *parent ) :
 polygonRoiToolBox::~polygonRoiToolBox()
 {
     delete viewEventFilter;
+    viewEventFilter = nullptr; // Will be tested in connect signal due to application closing
```
fcollot, line 140: It's not entirely clear how this can solve a connection-related bug. Which signal-slot connection is this referring to ?

src/plugins/legacy/polygonRoi/toolboxes/polygonRoiToolBox.cpp
```diff
@@ -191,10 +194,6 @@ void polygonRoiToolBox::updateView()
             return;
         }
         medAbstractImageView *v = qobject_cast<medAbstractImageView*>(view);
-        if (!v)
-        {
-            return;
-        }
```
fcollot, line 197: Removing these lines can only add a potential bug (if `updateView` is called on a view that is not a `medAbstractImageView` type), so it would be safer to keep them no ?

fcollot (changes requested):

src/plugins/legacy/polygonRoi/toolboxes/polygonRoiToolBox.cpp
fcollot: If something is calling `updateTableWidgetView` after the destructor has run, that is a bug that cannot be fixed this way. It is undefined behaviour to call a method on an object that has been destroyed.
(Normally Qt automatically disconnects objects that are destroyed, but this sort of bug can happen when Qt objects are deleted manually at the wrong time such as when there are pending signals in the event loop. Using `deleteLater` on Qt objects instead of `delete` avoids that).

fcollot: (I think my comment is misleading: I didn't mean to say that using `deleteLater`on `viewEventFilter`will correct the bug, because the problem is not the event filter but the toolbox itself that is being used after destruction)

mathildemerle: Thanks, i found a way to solve the problem. The PR is updated

src/plugins/legacy/polygonRoi/toolboxes/polygonRoiToolBox.cpp
```diff
@@ -432,32 +438,24 @@ void polygonRoiToolBox::updateTableWidgetView(unsigned int row, unsigned int col
...
-                    QList<medViewContainer*> containersInTab = tabs->containersInTab(tabs->currentIndex());
-                    if (containersInTab.size()==1 && containersInTab.at(0)->uuid()==mainContainerUUID)
-                    {
-                        if (activateTBButton->isEnabled())
-                        {
-                            repulsorTool->setEnabled(true);
-                            tableViewChooser->setEnabled(true);
-                        }
-                    }
-                }
```
juliencastelneau, line 457: I do not understand why you remove this lines.
In some scenario, we need to enable the repulsor and the tableView. You can not just remove it, no ?
Or maybe I missed something ?

mathildemerle: Maybe i was not doing the right tests, but i did not see any case where you need to enable the repulsor and orientation buttons at the closing of the alternate view. When you split the views with the orientation buttons, and you remove the alternate view, the buttons on the original view are good. Is there an other case where you would need it?

juliencastelneau: I can not remember rigth now.
But I do not write this code without reason. 
It was to answer to a particular scenario useful for users.
As I said before, I have to develop new features in this toolbow soon. 
I have planned to do a refactoring at this time.

juliencastelneau: Every crashs related to open/close view and/or containers must not be fixed in the toolbox.
It is a deeper issue in the way of managing the views and containers in core I think.

mathildemerle: I think i know the origin of this code: in order to avoid crashs, when the user choose to use the split button, the main view is blocked, meaning you can't remove it. If you remove the second view, the main view switch to the closable mode. However, before this PR you could remove the main view even with split views, meaning that at the removal of the first view you needed to use the alternate view as a new main view. This code is not needed anymore

juliencastelneau: ok but i made some actions on repulsor for instance
Are you sure the state of the repulsor is ok now ?

juliencastelneau: And It is important to keep the first view as a main view because the segmentation is made in this orientation.
Not sure, you can continue to segment the images without the first view.

juliencastelneau: I have read this post about C++ lambda expressions and Qt connect mechanism 
https://medium.com/genymobile/how-c-lambda-expressions-can-improve-your-qt-code-8cd524f4ed9f

This comments seems interesting:
" Qt connection system is smart enough to delete connections if either the sender or the receiver is deleted, so in our first version of setMonitor(), if monitor is deleted the connection is deleted as well. But now we use a lambda for the receiver: Qt has no way to now that this lambda makes use of monitor. Even if monitor is deleted, the lambda will still be called and the application will crash when it tries to dereference monitor.

To avoid that, you can pass a “context” argument to the connect() call, [...]
we pass monitor as a context to connect(). It won't affect the execution of our lambda, but when monitor is deleted, Qt will notice and will disconnect Worker::progress() from our lambda."

fcollot: Also, it is not useful to test `mainContainer` because it is captured by value in the lambda expression (see https://thispointer.com/c11-lambda-how-to-capture-local-variables-inside-lambda/), and you cannot capture it by reference since it is a local pointer.

mathildemerle: Great, if the main view should not change, it amplifies the need to lock the main view if the view is split.

I need to add that the lambda connection is done on `view` which is from `container->view()`, which is the split view, not the main one. So, this connection is only used at the closing on the second view. So, it means that at the closing of the second view, if there is only the main view, we want to activate the repulsor and split buttons. However, it is already done without this code.

I'm thinking about something else also: in order to avoid any problems, it would be interesting in the future not to allow the view to be split in the Tools button. Only the split buttons of the toolbox should be accepted i think.

Thanks for the documentation, it's very interesting. If i understand well, if we would like to keep the code removed here, or for a future code, we could pass a context variable. Here, i'm not sure which one, for instance `this->getWorkspace()` crash in it at the closing of the application. Would it be possible to pass `this` or a function?

fcollot: (I edited the previous comment because you can technically get a reference (by dereferencing), but the test would still be pointless since the reference will not change. What I meant was you shouldn't be testing `mainContainer`like this but instead call `containersInTabSelected.at(0)`directly inside the lambda if you want to check that the container still exists when the lambda is called)

mathildemerle: Oki, i'll remove the test on `mainContainer`. However, we could not test directly (without the use of context param) `containersInTabSelected` because `this->getWorkspace()->tabbedViewContainers();` crashes before that

fcollot: (our posts crossed) The context should be `this` and `mainContainer` should not be captured but accessed using `containersInTabSelected.at(0)`. That way it is not important who gets deleted first: if it is the toolbox (i.e. `this`) then the lambda will be disconnected, and if there is no container then you will get a null value (must test it of course).

mathildemerle: Thanks for your help, indeed it does not crash at the end of the application anymore, even with the call to `getWorkspace()`! What do you want to do with the lines:

```c++
                    if (containersInTab.size()==1 && containersInTab.at(0)->uuid()==mainContainerUUID)
                    {
                        if (activateTBButton->isEnabled())
                        {
                            repulsorTool->setEnabled(true);
                            tableViewChooser->setEnabled(true);
                        }
                    }
```
They seems not to be used anymore. We can remove them, or put them back and i can add a comment saying that they seems obsolete, or i can put them back. As you want

juliencastelneau: I prefer the second choice if possible.
I will work on it in few weeks

mathildemerle: Done :)

juliencastelneau (approved):

fcollot (approved):

PR #802, opened 2020-11-18
WIN32 - Copy dlls from external project to right place

juliencastelneau:
Creation of a new macro:
If a dll is generated in a plugin from an extenal project (MUSIC for instance), 
the dll poduced  is now copied to the right directory

PR #822, opened 2020-12-16
[Package] solve packaging pb on Ubuntu with .so files

mathildemerle:
Solve a packaging problem on Ubuntu due to https://github.com/medInria/medInria-public/pull/802 with .so files from medInria plugins which were removed before packaging.

:m:


Florent2305 (approved):
Ok, just verify if other changes will be needed to compile some external private plugins

PR #823, opened 2020-12-17
[Package] package fails because of patchs

mathildemerle:
The BUILD_ALWAYS at 0 created patch error at the packaging step.

:m:


Florent2305 (approved):
ok

PR #825, opened 2020-12-17
Handle workspace construction failure

fcollot:
This PR adds the possibility to handle workspace construction failure.

If an (unrecoverable) error is encountered in the constructor, it should throw an exception. This will be caught and handled by the workspace factory. The switch to the workspace will be cancelled and medInria will switch to the home area instead. To make sure resources are properly freed, a dedicated "parent" widget is given to the workspace so that all qwidgets it has created can be destroyed easily.


fcollot:
(I just pushed a new version of the commit to add a generic catch clause for non standard exceptions, to be more robust)

mjuhoor (approved):

paulineMig (approved):

PR #826, opened 2020-12-17
Allow non-standard workspace registration

fcollot:
The non-templated version of registerWorkspace was protected. This PR makes it public so that workspaces can register their own creator.

(GitHub is not displaying the changes correctly. The only code that moved is the registerWorkspace method)

### DIFF
--- a/src/layers/legacy/medCoreLegacy/gui/factories/medWorkspaceFactory.h
+++ b/src/layers/legacy/medCoreLegacy/gui/factories/medWorkspaceFactory.h
@@ -90,21 +90,6 @@ public:
                                  isActive);
     }
 
-
-    QHash<QString, medWorkspaceFactory::Details *> workspaceDetails() const;
-    QList<medWorkspaceFactory::Details *> workspaceDetailsSortedByName(bool activeOnly = false) const;
-    medWorkspaceFactory::Details * workspaceDetailsFromId(QString identifier) const;
-
-    bool isUsable(QString identifier) const;
-
-    QList<QString> workspacesFromCategory(const QString& category) const;
-
-public slots:
-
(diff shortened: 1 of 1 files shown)


fcollot:
(for info this is all used for the pipeline workspace)

mjuhoor (approved):

src/layers/legacy/medCoreLegacy/gui/factories/medWorkspaceFactory.h
```diff
@@ -124,6 +109,21 @@ public slots:
...
+    QList<medWorkspaceFactory::Details *> workspaceDetailsSortedByName(bool activeOnly = false) const;
+    medWorkspaceFactory::Details * workspaceDetailsFromId(QString identifier) const;
+
+    bool isUsable(QString identifier) const;
+
+    QList<QString> workspacesFromCategory(const QString& category) const;
+
+public slots:
+
+    medAbstractWorkspaceLegacy* createWorkspace(QString type,QWidget* parent=0);
```
paulineMig, line 123: Since you are modifying the fill you could put parent to nullptr instead of 0

paulineMig (approved):

src/layers/legacy/medCoreLegacy/gui/factories/medWorkspaceFactory.h
fcollot: github is displaying this all wrong. I didn't touch these lines of code, I moved the function that was below to put it above.
Also I know it would be a minimal change but by principal I don't like changing code that isn't the target of the PR.

juliencastelneau (approved):

PR #827, opened 2021-01-07
[Packaging] solve external projects plugins and libs problems

mathildemerle:
From https://github.com/Inria-Asclepios/medInria-public/pull/687

This PR solves problems with external project packaging: lib not at the right place, duplication of plugins, etc.
For now it has been tested on Ubuntu 20.

This has been developed with @Florent2305.

:m:


mjuhoor:
Cool. I'll approve when it's tested on mac (and windows ?)

Florent2305 (approved):
I am biased because I myself participated very greatly in this PR.
But it has been tested on several different platforms with some success, so I approve of it.

PR #834, opened 2021-01-13
[StatsROI] useless call to empty()

mathildemerle:
From a compilation from scratch i saw a warning `ignoring return value of 'bool std::vector<_Tp, _Alloc>::empty() const [with _Tp = double; _Alloc = std::allocator<double>]', declared with attribute nodiscard [-Wunused-result]`. 

The call to `this->computedOutput.empty();` is wrong because it does not clear the vector (we should use `clear()` instead), it only tests if the vector is empty or not. However, here, it is not needed to test if the vector is empty, or to clear it, so i removed the line.

:m:

### DIFF
--- a/src/layers/legacy/medUtilities/statsROI.cpp
+++ b/src/layers/legacy/medUtilities/statsROI.cpp
@@ -2,12 +2,14 @@
 
  medInria
 
- Copyright (c) INRIA 2013 - 2020. All rights reserved.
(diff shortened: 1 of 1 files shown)


fcollot (approved):

mathildemerle:
This PR has been rebased

Florent2305 (approved):
Ok

PR #836, opened 2021-01-14
set_lib_install_rules: remove useless headers_list

fcollot:
The `headers_list` argument is not used and prevents calling this macro on libs that do not have associated headers (which is the case for bindings libraries).

### DIFF
--- a/src/cmake/module/set_lib_install_rules.cmake
+++ b/src/cmake/module/set_lib_install_rules.cmake
@@ -11,7 +11,7 @@
 #
 ################################################################################
 
-macro(set_lib_install_rules_generic target dest headers_list)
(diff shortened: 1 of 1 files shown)


mathildemerle:
We used like this the headers on external libs: https://github.com/Inria-Asclepios/music/pull/826

fcollot:
The parameter is not used in the macro code. It is defined as a required parameter but the macro ignores it and accesses the headers list using ARGV (which contains all the arguments). Removing the parameter has no effect on any invocation of the macro (but has the bonus side effect of allowing us to use ARGN to access the headers: see https://cmake.org/cmake/help/latest/command/macro.html)

mathildemerle (approved):

PR #841, opened 2021-02-05
[medPython] Initialization

fcollot:
Initializes the Python interpreter at the start of the application. For now this uses the Python interpreter installed on the system, which must be Python 3 (and ideally 3.9). You need the Python 3 development headers to compile, and the application will crash if the executable is run without Python 3 installed. At a later time I'll look into how we include the interpreter in the package.
_(This component is mostly empty for now but the upcoming PRs will add to it)_

### DIFF
--- a/src/layers/medPython/cmake/python_utils.cmake
+++ b/src/layers/medPython/cmake/python_utils.cmake
@@ -1,0 +1,17 @@
+################################################################################
+#
+# medInria
+
+# Copyright (c) INRIA 2021. All rights reserved.
+
+# See LICENSE.txt for details in the root of the sources or:
+# https://github.com/medInria/medInria-public/blob/master/LICENSE.txt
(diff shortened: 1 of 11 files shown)


src/layers/medPython/cmake/python_utils.cmake
```diff
@@ -0,0 +1,17 @@
...
+
+# See LICENSE.txt for details in the root of the sources or:
+# https://github.com/medInria/medInria-public/blob/master/LICENSE.txt
+
+# This software is distributed WITHOUT ANY WARRANTY; without even
+# the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+# PURPOSE.
+#
+################################################################################
+
```
fcollot, line 15: There will be more code here later.

PR #847, opened 2021-02-16
[iterativeClosestPoint] Add Export Transformation Matrix Option

aarrieul:
Add an option to get the transformation matrix of an iterative closest point process. 
Use it to transfer points from a patient to another.

### DIFF
--- a/src/plugins/legacy/iterativeClosestPoint/iterativeClosestPointProcess.cpp
+++ b/src/plugins/legacy/iterativeClosestPoint/iterativeClosestPointProcess.cpp
@@ -27,6 +27,7 @@
 #include <vtkMetaDataSet.h>
 #include <vtkMetaSurfaceMesh.h>
 #include <vtkSmartPointer.h>
+#include <vtkMatrix4x4.h>
 
 // /////////////////////////////////////////////////////////////////
 // iterativeClosestPointProcessPrivate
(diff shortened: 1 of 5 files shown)


mathildemerle (approved):

src/plugins/legacy/iterativeClosestPoint/iterativeClosestPointProcess.cpp
```diff
@@ -184,6 +220,36 @@ int iterativeClosestPointProcess::update()
     return medAbstractProcessLegacy::SUCCESS;
 }
 
+
+void iterativeClosestPointProcess::exportTransformMatrix()
+{
+    if (d->exportMatrixState == 2)
+    {
+        QByteArray matrixStr;
+        matrixStr += "# Transformation matrix from " + d->sourceName + " to " + d->targetName + "\n";
```
paulineMig, line 229: Because you are adding a comment as the first line your export file is not usable as it in meshManipulation toolbox. You need to modify  importTransform function of this toolbox so it can skip commented lines

aarrieul: Problem solved

paulineMig (approved):

Florent2305 (approved):
ok

Issue #850, opened 2021-03-03
4D navigator problem, missing one frame


ocommowi:
On version 3.2, I have experienced a weird behavior. Load any 4D volume, and you will see that there is one less frame than you would expect in the navigator, e.g. if you have 3 3D volumes in your 4D volume, the number of frames will be 2. 

Thus there is at least one volume you will not see (for my example the middle one). The time is also not respecting the real time: all in all it seems that everything works except the count of frames is wrong

mathildemerle:
Could be linked to https://github.com/medInria/medInria-public/pull/815

Florent2305:
> Could be linked to #815

Indeed

Florent2305:
As suggested by @mathildemerle, when troubleshooting this issue we can see if it is related with https://github.com/medInria/medInria-public/issues/845

paulineMig:
By bad i totally forgot to test 4D images but i know where is the issue i will do a fix

PR #851, opened 2021-03-04
[medVtkViewItkDataImage4DInteractor]: fix wrong frameRateComputation

paulineMig:
fix https://github.com/medInria/medInria-public/issues/850

### DIFF
--- a/src/plugins/legacy/itkDataImage/interactors/medVtkViewItkDataImage4DInteractor.cpp
+++ b/src/plugins/legacy/itkDataImage/interactors/medVtkViewItkDataImage4DInteractor.cpp
@@ -122,11 +122,12 @@ void medVtkViewItkDataImage4DInteractor::setInputData(medAbstractData *data)
 
         if (SetViewInput(data, layer) )
         {
+            double frameRate = static_cast<double>(m_poConv->getNumberOfVolumes()) / static_cast<double>(m_poConv->getTotalTime());
             d->imageData->setMetaData("SequenceDuration", QString::number(m_poConv->getTotalTime()));
-            d->imageData->setMetaData("SequenceFrameRate", QString::number((double)(m_poConv->getNumberOfVolumes() -1 )/ (double)m_poConv->getTotalTime()));
+            d->imageData->setMetaData("SequenceFrameRate", QString::number(frameRate));
 
             qDebug() << "SequenceDuration" << m_poConv->getTotalTime();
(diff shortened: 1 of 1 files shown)


mathildemerle (approved):

src/plugins/legacy/itkDataImage/interactors/medVtkViewItkDataImage4DInteractor.cpp
```diff
@@ -122,11 +122,12 @@ void medVtkViewItkDataImage4DInteractor::setInputData(medAbstractData *data)
 
         if (SetViewInput(data, layer) )
         {
+            double frameRate = static_cast<double>(m_poConv->getNumberOfVolumes()) / static_cast<double>(m_poConv->getTotalTime());
```
fcollot, line 125: Just for info, there is no need to cast both operands to double. Only one is sufficient (a division involving a double results in a double).

fcollot (approved):

src/plugins/legacy/itkDataImage/interactors/medVtkViewItkDataImage4DInteractor.cpp
paulineMig: True i just did a copy paste without thinking do you want me to modify it?

fcollot: no don't worry it's just something I see often in the code so I wanted to point it out to

mjuhoor (approved):

src/plugins/legacy/itkDataImage/interactors/medVtkViewItkDataImage4DInteractor.cpp
```diff
@@ -122,11 +122,12 @@ void medVtkViewItkDataImage4DInteractor::setInputData(medAbstractData *data)
...
             d->imageData->setMetaData("SequenceDuration", QString::number(m_poConv->getTotalTime()));
-            d->imageData->setMetaData("SequenceFrameRate", QString::number((double)(m_poConv->getNumberOfVolumes() -1 )/ (double)m_poConv->getTotalTime()));
```
nebatmusic, line 126: What is the frame rate. If it is the number of volumes per unit of time, you should keep it as it was.

mathildemerle: framerate is now computed line 125 with this same computation except the "-1"

PR #861, opened 2021-03-18
[medDatabaseExporter] Fix problem when exporting big data

mjuhoor:
- Fix problem when exporting very big medAbstractData onto disk: the object would be garbage collected before the writing would finish. Correct usage of smart pointers fixes this.


mathildemerle (approved):

Florent2305 (approved):
Good :)

PR #865, opened 2021-04-08
[medExternalResources] bundle object should not be freed

fcollot:
(based on https://github.com/medInria/medInria-public/pull/863)
commit: https://github.com/medInria/medInria-public/commit/e173ea6550c7158f4eb3c0544f03c9071e992a41

Fixes a crash that occurs when retrieving more than one external resource for a single target.


Florent2305 (approved):
It's ok

mathildemerle (approved):

PR #871, opened 2021-05-05
[medDatabase] Fix suffix added to series

### LINKED ISSUE #820: medDatabaseImporter::ensureUniqueSeriesName prevents duplicate series names even between unrelated studies or patients
When importing new data in `medAbstractDatabaseImporter::importData`, the function `medDatabaseImporter::ensureUniqueSeriesName` is called. In that function, if the series name of the new data already exists anywhere in the database, then a prefix is added to it. This constraint should have been limited to the same study.

This is an issue with `medAbstractDatabaseImporter::importData`, not `medAbstractDatabaseImporter::importFile`, and so will only appear when data is imported from within medInria (output of a toolbox for example). You can reproduce this by creating a new data inside medInria (from any toolbox). While it is still a temporary data, rename it to a series name that already exists in the DB (from a different patient). Now save the new data, and it will have a prefix added to it.

mjuhoor:
- Fixes #820
- SQL request is restricted to the study if possible

### DIFF
--- a/src/layers/legacy/medCoreLegacy/database/medAbstractDatabaseImporter.cpp
+++ b/src/layers/legacy/medCoreLegacy/database/medAbstractDatabaseImporter.cpp
@@ -429,11 +429,6 @@ void medAbstractDatabaseImporter::importData()
         return;
     }
 
-    // Update name of the series if a permanent data has this name already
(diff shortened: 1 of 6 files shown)


fcollot (approved):
Thanks !

juliencastelneau (approved):

PR #881, opened 2021-06-03
[add_external_resources] remove RCC option, use POST_BUILD

fcollot:
Following discussion with @Florent2305, we decided to make the following changes to the `add_external_resources` macro:

- Change the custom command that copies the files so that it executes at post build time.
- Remove the RCC option. It is not necessary and creates some complication with the other changes.

### DIFF
--- a/src/cmake/module/add_external_resources.cmake
+++ b/src/cmake/module/add_external_resources.cmake
@@ -15,36 +15,21 @@ macro(add_external_resources target)
 
 ################################################################################
 #
-# Usage: add_external_resources(target [RCC name] [PREFIX prefix] resources...)
+# Usage: add_external_resources(target resources...)
 #
 # External resources will he handled as follows:
 #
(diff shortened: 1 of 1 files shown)


src/cmake/module/add_external_resources.cmake
```diff
@@ -135,16 +72,21 @@ macro(add_external_resources target)
...
-        string(REPLACE " " "_" target_suffix "${filename}")
-        add_custom_target(${target}_resource_${target_suffix} ALL
-          DEPENDS "${resource}" "${output_file}")
+        list(APPEND copied_resources "${output_file}")
       endforeach()
 
+      set(resources "${copied_resources}")
+
     endif()
+
```
fcollot, line 85: The lines below have not changed. They are just moved here for clarity (the above code handles build rules, the code below handles packaging, so it is more logical in this order).

juliencastelneau (approved):

paulineMig (approved):

PR #886, opened 2021-06-22
[medDoubleParameterL]: fix wrong (0,0) shown values when attributes has only one value for all mesh

paulineMig:
- open test_mesh.vtk go to attributes Zones -> min and max intensity are at 0 but they should not the true value is 1
[test_mesh.zip](https://github.com/medInria/medInria-public/files/6692796/test_mesh.zip)


mathildemerle (approved):

Florent2305 (approved):
Ok

PR #890, opened 2021-07-05
[medInria3.2] bad version number

mathildemerle:
Correct wrong version number seen on the binary package.

:m:


Florent2305 (approved):
ok

PR #894, opened 2021-07-09
[medPython] Lazy loading

fcollot:
Introduces lazy loading of Python (which was asked by @Florent2305 ). The Python library is now loaded and initialized when the wrapping API is first used.
(this change removes the need for the **--no-python** option)

### DIFF
--- a/src/app/medInria/main.cpp
+++ b/src/app/medInria/main.cpp
@@ -29,10 +29,6 @@
 #include <medSettingsManager.h>
 #include <medStorage.h>
 
-#if defined(USE_PYTHON)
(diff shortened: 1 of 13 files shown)


mathildemerle (approved):

fcollot:
Pushed the following changes:
- optimised lazy loading: the test is only done on object creation, not every time a function is called
- improved wrapping templates of core functions to simplify the calling syntax
- removed `getPythonPath` as it does not return the correct result if the path has been changed.

juliencastelneau (approved):

Issue #868, opened 2021-04-16
[License] List licences of third-parties libraries and their sources 


mathildemerle:
The licenses of external libraries that medInria uses have some requirements, such as list the licenses and the sources in the software/website.

Done in https://github.com/Inria-Asclepios/medInria-public/pull/704 that i could adapt (i just need to remove the unused lib in medInria).

This LICENSES_EXT.txt file could be used on the medInria website too.

PR #908, opened 2021-09-08
[License] external libs list with licenses and sources links

mathildemerle:
Fix https://github.com/medInria/medInria-public/issues/868

Most licenses ask to be listed in the application. This Pull Request adds an External Licenses action in the application menu.

:m:


Florent2305 (approved):
Ok but will change with rebranding mechanism

PR #913, opened 2021-09-10
Allow framework plugins and resources

fcollot:
I modified `set_plugin_install_rules` to allow using `add_external_resources` on plugins. When using this macro on macOS the library becomes a framework. This causes an issue when loading because QPluginLoader complains it can't find the shared library. To fix this we need to replace the framework path with the path to the library inside the framework (I don't know why QPluginLoader doesn't do this automatically).


Florent2305 (approved):
It's ok

Issue #912, opened 2021-09-09
Registration workspace open icon not working with empty view


ocommowi:
In the registration workspace, when starting from blank empty views, the icons on top of the vies do not work. So there is no way to load an image from the views as in the other workspaces. It's a bit problematic to have to go back to the browser to load the images.

PR #917, opened 2021-09-14
[viewContainer] background and buttons by default without data

mathildemerle:
This PR:

 * fix https://github.com/medInria/medInria-public/issues/633 where we were loosing the default background color/buttons after removing a data in a view in fix containers workspace (Filtering, Registration for instance).
 * fix https://github.com/medInria/medInria-public/issues/912 where the Tools button were locked on an empty view in the same fix containers.
 * fix the "Unable to retrieve data at layer: 0 from: \"medVtkView\"" warning after removing a data

PS: this PR has been made on master branch since medInria3.2 was temporary not compiling on my side due to a change in pre-requisites, but these fixes are really interesting for med3.2 aswell.
 
:m:

### DIFF
--- a/src/layers/legacy/medCoreLegacy/views/medAbstractLayeredView.cpp
+++ b/src/layers/legacy/medCoreLegacy/views/medAbstractLayeredView.cpp
@@ -238,7 +238,6 @@ void medAbstractLayeredView::removeData(medAbstractData *data)
 
     if( res > 0 )
     {
-        emit layerRemoved(layer);
         emit layerRemoved(data);
     }
 }
--- a/src/layers/medWidgets/process/arithmetic_operation/medAbstractArithmeticOperationProcessPresenter.cpp
+++ b/src/layers/medWidgets/process/arithmetic_operation/medAbstractArithmeticOperationProcessPresenter.cpp
@@ -72,17 +72,31 @@ medViewContainerSplitter *medAbstractArithmeticOperationProcessPresenter::buildV
     medViewContainer * outputContainer = inputContainer1->splitVertically();
     medViewContainer* inputContainer2 = inputContainer1->splitHorizontally();
 
-    inputContainer1->setDefaultWidget(new QLabel("INPUT 1"));
+    auto viewLayout1 = static_cast<QVBoxLayout*>(inputContainer1->defaultWidget()->layout());
+    auto label1 = static_cast<QLabel*>(viewLayout1->itemAt(0)->widget());
+    if (!label1->text().contains("INPUT 1"))
+    {
+        label1->setText("INPUT 1\n\n"+label1->text());
+        label1->setAlignment(Qt::AlignCenter);
+    }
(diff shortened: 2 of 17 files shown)


src/layers/legacy/medCoreLegacy/views/medAbstractLayeredView.cpp
```diff
@@ -238,7 +238,6 @@ void medAbstractLayeredView::removeData(medAbstractData *data)
 
     if( res > 0 )
     {
-        emit layerRemoved(layer);
```
mathildemerle, line 241: connected to a method witch try working on a delete data, creating the previous "Unable to retrieve data at layer: 0 from: "medVtkView"" warning. Line not needed

src/layers/medWidgets/process/arithmetic_operation/medAbstractArithmeticOperationProcessPresenter.cpp
```diff
@@ -72,17 +72,31 @@ medViewContainerSplitter *medAbstractArithmeticOperationProcessPresenter::buildV
     medViewContainer * outputContainer = inputContainer1->splitVertically();
     medViewContainer* inputContainer2 = inputContainer1->splitHorizontally();
 
-    inputContainer1->setDefaultWidget(new QLabel("INPUT 1"));
+    auto viewLayout1 = static_cast<QVBoxLayout*>(inputContainer1->defaultWidget()->layout());
```
mathildemerle, line 75: In a fix container that you can't remove (for instance in Filtering or Registration), you now can open a data from the buttons in the view or the Tools button. We still can add a label on the fix view, as done here. It will be displayed on top of the view buttons

src/layers/medWidgets/process/arithmetic_operation/medAbstractArithmeticOperationProcessPresenter.cpp
```diff
@@ -72,17 +72,31 @@ medViewContainerSplitter *medAbstractArithmeticOperationProcessPresenter::buildV
...
+        label2->setAlignment(Qt::AlignCenter);
+    }
     inputContainer2->setClosingMode(medViewContainer::CLOSE_VIEW);
     inputContainer2->setUserSplittable(false);
     inputContainer2->setMultiLayered(false);
 
-    outputContainer->setDefaultWidget(new QLabel("OUTPUT"));
+    auto viewLayoutOutput = static_cast<QVBoxLayout*>(outputContainer->defaultWidget()->layout());
+    auto labelOutput = static_cast<QLabel*>(viewLayoutOutput->itemAt(0)->widget());
+    labelOutput->setText("OUTPUT");
```
mathildemerle, line 99: on fix containers with `setUserOpenable(false)` the buttons are hidden/deactivated, so we can change the label directly

Florent2305 (approved):
Ok

PR #918, opened 2021-09-15
Optional argument to workspace and toolbox creator functions

fcollot:
The factory classes use functions pointers to create instances. SWIG cannot map Python functions to C function pointers, so it is not possible to register types from Python code the way it is done on the C++ side. As a workaround I created generic creator functions that take a pointer to a Python class and create an instance of that class. In order for that to work we need to add an (optional) argument to the creator functions.
(These changes should be extended to all the factories later on. I only did the ones needed for the pipeline plugin)

### DIFF
--- a/src/layers/legacy/medCoreLegacy/gui/factories/medToolBoxFactory.h
+++ b/src/layers/legacy/medCoreLegacy/gui/factories/medToolBoxFactory.h
@@ -47,7 +47,7 @@ class MEDCORELEGACY_EXPORT medToolBoxFactory : public dtkAbstractFactory
     Q_OBJECT
 
 public:
-    typedef medToolBox *(*medToolBoxCreator)(QWidget *parent);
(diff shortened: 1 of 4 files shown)


src/layers/legacy/medCoreLegacy/gui/factories/medToolBoxFactory.h
```diff
@@ -69,9 +69,16 @@ class MEDCORELEGACY_EXPORT medToolBoxFactory : public dtkAbstractFactory
                                toolboxType::staticName(),
                                toolboxType::staticDescription(),
                                toolboxType::staticCategories(),
-                               creator);
+                               creator,
+                               nullptr);
     }
```
fcollot, line 75: The non-templated registerToolBox function below is moved to the public section to be accessible from Python.

mathildemerle (approved):

Florent2305 (approved):
Good

PR #923, opened 2021-09-23
[VariationalSeg] solve crash avec 3D->2D view segmentation

### LINKED ISSUE #919: [VariationalSeg] previously 3D view crash in VariationalSeg
How to get the crash:

 * Open a data in Voi Cutter
 * Draw a shape and "Enter"
 * Switch to Variational Segmentation
 * Do 3 shift-clicks to create landmarks
 * Apply segmentation: crash "Erreur de segmentation (core dumped)"

mathildemerle:
Fix #919

This PR:
 * Change the display of the output at the end of the process in `VarSegToolBox::displayOutput()`, which was crashing with a 3D/2D view
 * Remove 2 non-used variable (mprOn was always set to false, and originalView not used if mprOn is set at false)
 * The `d->currentView->property("Orientation")` line created a warning since there are no Orientation property here
 * Handle `updateView` if the view is empty
 * The explanation text is smaller because it created a small graphic bug with the information button

:m:

### DIFF
--- a/src/plugins/legacy/variationalSegmentation/varSegToolBox.cpp
+++ b/src/plugins/legacy/variationalSegmentation/varSegToolBox.cpp
@@ -24,6 +24,7 @@
 #include <medRunnableProcess.h>
 #include <medTabbedViewContainers.h>
 #include <medUtilities.h>
+#include <medViewContainer.h>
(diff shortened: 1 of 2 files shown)


Florent2305 (approved):
ok

PR #928, opened 2021-09-24
Removed selection toolbox, refactored toolbox addition/removal

fcollot:
This PR makes changes to improve the way toolboxes are added and removed from workspaces, and to give more control over the default toolboxes:

- Remove the "selection" toolbox which serves no purpose except to group the toolboxes into one parent toolbox.
- Place the progression stack widget into its own toolbox.
- Add getters for the default toolboxes (mouse interaction, navigator, layers, progression stack)
- Remove unused methods to add and remove toolboxes in `medBrowserArea`, `medWorkspaceArea` and `medToolBox`, and keep only the ones in `medAbstractWorkspaceLegacy` (and add an `insertToolBox` method to mirror the one in `medToolBoxContainer`).
- Use signal/slot mechanics to connect the methods to `medToolBoxContainer`.
- Properly disconnect the workspaces from all signals when switching workspaces.
- Remove `clear` and `clearWorkspaceToolBoxes` (not used and not needed)

### DIFF
--- a/src/layers/legacy/medCoreLegacy/gui/medAbstractWorkspaceLegacy.cpp
+++ b/src/layers/legacy/medCoreLegacy/gui/medAbstractWorkspaceLegacy.cpp
@@ -49,12 +49,12 @@ public:
     QHash <QListWidgetItem*, QUuid> containerForLayerWidgetsItem;
 
     QList<medToolBox*> toolBoxes;
-    medToolBox *selectionToolBox;
     medToolBox *layersToolBox;
     medToolBox *layerListToolBox;
     medToolBox *interactorToolBox;
     medToolBox *navigatorToolBox;
     medToolBox *mouseInteractionToolBox;
+    medToolBox* progressionStackToolBox;
     medListWidget* layerListWidget;
 
     QList<QListWidgetItem*> selectedLayers;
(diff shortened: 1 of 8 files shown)


fcollot:
(this has no impact on the current workspaces (unless I've introduced a bug of course) but is required for the upcoming pipeline workspace)

src/layers/legacy/medCoreLegacy/gui/medAbstractWorkspaceLegacy.cpp
```diff
@@ -171,13 +170,26 @@ medTabbedViewContainers* medAbstractWorkspaceLegacy::tabbedViewContainers() cons
     return d->viewContainerStack;
 }
 
-void medAbstractWorkspaceLegacy::clear()
```
mathildemerle, line 174: it was not used when the user switches between a workspace to another?

mathildemerle (approved):

src/layers/legacy/medCoreLegacy/gui/medAbstractWorkspaceLegacy.cpp
fcollot: no

mjuhoor (approved):

PR #929, opened 2021-09-24
add "shallow" loading option when retrieving data

fcollot:
_(edited after refactoring)_

This adds the possibility to get information on the type of a data in the database. This is done by adding a `medDataManager::getDataType` method that returns the `QMetaObject` instance of the data class. In order for this to work I have added an option in the DB loading mechanism that retrieves empty `medAbstractData` instances (the actual data is not loaded, only the `dtkAbstractDataReader::readInformation` method is used to obtain the instance.

### DIFF
--- a/src/layers/legacy/medCoreLegacy/database/medAbstractDbController.h
+++ b/src/layers/legacy/medCoreLegacy/database/medAbstractDbController.h
@@ -48,7 +48,7 @@ signals:
     void dataRemoved(const medDataIndex& index);
 
 public slots:
-    virtual medAbstractData* retrieve(const medDataIndex& index, bool fullData) const = 0;
+    virtual medAbstractData* retrieve(const medDataIndex& index, bool readFullData = true) const = 0;
 
     virtual void importData(medAbstractData* data, const QUuid& importUuid) = 0;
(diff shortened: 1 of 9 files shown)


mathildemerle (dismissed):

mjuhoor (dismissed):

fcollot:
I have refactored this coded (and updated the header comment) following discussion with @Florent2305 who did not think it was a good idea to be able to retrieve empty `medAbstractData` instances (the data pointer within them is void* so will cause crashes if used like normal data). My intention was to use this option to know what will be the actual class of the loaded data, and so I figured a better way to use this code by returning the `QMetaObject` instance instead of the incomplete `medAbstractData` instance. Using this you can test if a data is a mesh for example by using the following code:
```
if (medDataManager::instance().getDataType(index).inherits(medAbstractMeshData::staticMetaObject))
   // etc.
```

fcollot:
(I need new reviews on this since the code and functionality has changed)

mathildemerle (approved):

juliencastelneau (approved):

Florent2305 (approved):
It's Ok

PR #930, opened 2021-09-27
[medStatusBar] removal of wrong message

fcollot:
`medStatusBar::removeMessage` assumes the message being removed is the first on the list. This is not always the case and causes a crash when it happens because the deleted message is still on the stack.


mjuhoor:
Thank you, this was really a nasty one.

mjuhoor (approved):

mathildemerle (approved):

Florent2305 (approved):

Issue #934, opened 2021-10-18
[Documentation] enhancement for medInria and its plugins shop


mathildemerle:
Due to the future changes of medInria 4 with the plugin shops, it could be interesting to enhance the documentation system.

Here are some suggestions after a brainstorming in the MUSICardio team that could be added to [Objectives 4.x](https://github.com/medInria/medInria-public/issues/864) list:

### Help button for each toolbox
The Help button (clicked in toolbox header) should be dedicated to a toolbox, and not a plugin (which can contain multiple toolboxes). 
It allows creators of plugins (and external plugins for instance) to create their own documentation for their plugins without having to change the medInria documentation or create a global "fork" of the documentation.
It means that we can decide:
 * to only have one toolbox per plugin
 * to change the way the help pop-up is created (right now it's filled with the information in `XXXPlugin::description()` for instance)

### Help button with more place for info
Right now, the area of help information in the Help button is really small:
![helpNow](https://user-images.githubusercontent.com/3910352/137751070-8680d926-4ac5-4e8d-9916-6db2ba76f7db.png)

This tool should be enhanced to get a wider area for the description, as:
![helpNew](https://user-images.githubusercontent.com/3910352/137753138-0ab298ac-f646-454b-9921-d385bd30cb76.png)

### Help info with images
If it's possible, add possibility to add screenshots for instance or fancy way to create a nice documentation.

### Help button in markdown
As said in https://github.com/medInria/medInria-public/issues/867, after Qt 5.14, we would be able to write documentation in markdown in QTextEdit. It will be easier to create nice documentation. Right now it's in html.

### Global medInria documentation without toolboxes
The global medInria documentation could be written collaboratively to be easier to maintain. And it could focus on the core of medInria, and not the toolboxes, since the toolboxes documentation could be written in plugins.

### Global medInria documentation includes the wiki info
In a collaborative way also, it could be interesting to add the wiki compilation info pages (and their [suggestion of enhancement](https://github.com/medInria/medInria-public/issues/891) to the global medInria documentation.

PS : the actual documentation widget is a dtkAboutPlugin, and since we want to remove dtk from the project, it's a step in this direction

PR #936, opened 2021-10-20
[Documentation] add new doc widget for each plugin/tlbx

mathildemerle:
Part of this issue: https://github.com/medInria/medInria-public/issues/934

It:
 * create a new documentation widget instead of dtkAboutPlugin
 * leave more space for documentation
 * possibility to add html text: text/images/headers/etc
 * create the documentation for the Segmentation toolboxes for now
 * remove medButton which was only used for the about/help button and was a mimic of QPushButton
 * the help button is on the same line as the selector of toolbox, which is more clear for user that they can click on it to have info
 * the help button is a new one
 * the authors and version are written on the help window if there are not empty

Example:

screenPopup

:m:


Florent2305 (changes requested):
Globally ok but please take into account request changes

mathildemerle:
This PR is rebased and the asked macro has been added

Florent2305 (approved):
ok

PR #940, opened 2021-10-25
[Reslice] fix crash with spacing/dim parameters

mathildemerle:
Fix a crash when you switch the Spacing/Dimension parameters. 

Appears after https://github.com/Inria-Asclepios/medInria-public/pull/718 because one of the updateView is empty at the end.

:m:


Florent2305 (approved):
ok

PR #941, opened 2021-10-25
[Reslice] fix crash with spacing/dim parameters - 3.2

mathildemerle:
Same as https://github.com/medInria/medInria-public/pull/940 on medInria3.2

Fix a crash in reslice switching the spacing/dimension parameters.

:m:


Florent2305 (approved):
Ok

PR #954, opened 2022-01-06
Fix mesh readers

paulineMig:
- In volume meshes, edges, triangles, ridges are read and not only tetrahedra as before.
- Similarly in surface meshes, edges, ridges are read and not only triangles.
- Add vtu reader and writer for mesh of type unstructured grid
- Prevent use of vtp writer for unstructured grid and vtu writer for polydata
- Check the version of the vtk file to avoid crash due to the new format introduced in vtk 9

### DIFF
--- a/src/layers/legacy/medVtkDataMeshBase/vtkMetaSurfaceMesh.cxx
+++ b/src/layers/legacy/medVtkDataMeshBase/vtkMetaSurfaceMesh.cxx
@@ -281,14 +281,19 @@ void vtkMetaSurfaceMesh::Write (const char* filename)
 {
     try
     {
-        qDebug()<<"Writing: "<<filename;
         if (vtkMetaSurfaceMesh::IsVtpExtension(vtksys::SystemTools::GetFilenameLastExtension(filename).c_str()))
         {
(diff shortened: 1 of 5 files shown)


src/layers/legacy/medVtkDataMeshBase/vtkMetaSurfaceMesh.cxx
```diff
@@ -345,192 +364,219 @@ unsigned int vtkMetaSurfaceMesh::CanReadFile (const char* filename)
...
+    if(positionStream(fileInput, elementsNameHash[elementsNames[3]].first+1, line) &&
+            elementsNameHash[elementsNames[1]].second > 0)
+    {
+        curLine = 0;
+        while (getline(fileInput, line) && curLine < elementsNameHash[elementsNames[3]].second)
+        {
+            std::istringstream  lineStream(line);
+            unsigned int pos;
+            lineStream >> pos;
+            //This work because the edges were the first cells added
```
mathildemerle, line 553: *works

mathildemerle (approved):

Florent2305 (approved):
It's good for me

PR #959, opened 2022-01-31
[medToolBox] component and value accessors

fcollot:
Aside from some minor python fixes, this PR adds a couple important functions in `medToolBox`. These are changes I made a while back to replace the hacks we made in the old pipeline code when accessing and modifying toolbox widgets (to set values, click buttons etc.). The following virtual functions are added:
- `getComponent`: used to click buttons for example
- `getValue` and `setValue`: used to access things like spinbox values.

The old pipeline code used `QObject::findChild` to achieve this, and the default implementation of these functions does the same. The old code also had parts that directly accessed some special functions in the histogram toolbox (which meant we had to link with that plugin). This is no longer possible because the pipeline code is in the core, so I have modified the histogram toolbox so that it defines it's own implementation of the above functions (not posted yet).

I initially wrote this code even before we were talking about medInria 3 (so yes, a long time ago), as a suggestion to improve the architecture of medInria 2 to make it more adapted to the pipelines. I never ended up posting it but these changes are necessary for medInria 3 specifically because of the histogram issue (and also I had already adapted the pipeline code in the upcoming version to these changes). _**These are only temporary aditions for medInria 3, they are of course not intended for medInria 4 which has an entirely different toolbox philosophy.**_

### DIFF
--- a/src/layers/legacy/medCoreLegacy/gui/toolboxes/medToolBox.cpp
+++ b/src/layers/legacy/medCoreLegacy/gui/toolboxes/medToolBox.cpp
@@ -13,6 +13,7 @@
 
 =========================================================================*/
 
+#include <medAbstractParameterL.h>
 #include <medAbstractProcessLegacy.h>
 #include <medButton.h>
 #include <medJobManagerL.h>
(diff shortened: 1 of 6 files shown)


mathildemerle (approved):

Florent2305 (approved):
ok

PR #961, opened 2022-02-03
Plugin aphp pacs restapi

juliencastelneau:
data source plugin for APHP purpose (ADT medtag)
To work, the build must use a specific branch of QtDCM julcast/PluginAPHP which is a draft version for now
The goal is to test the importation/exportation of data from/to a PACS in our new dataSource plugin mechanism


Florent2305 (approved):
Ok

PR #962, opened 2022-02-11
[medPython] settings to allow adding python paths (for python plugins)

fcollot:
Adds a settings value that contains additional python paths to append to the python modules path at startup.
A python settings widget is also added to allow the user to add/remove paths.


Florent2305 (approved):
ok

PR #963, opened 2022-02-11
[medPythonTools] improve bindings and add QHash

fcollot:
- adds bindings for QHash
- corrects bindings for `medAbstractData` smart pointers
- allows binding of output arguments

PR #964, opened 2022-02-11
add medPipeline (work in progress)

fcollot:
(uses https://github.com/medInria/medInria-public/pull/963)
(adds commit https://github.com/medInria/medInria-public/commit/c16f3c1e83066c7b94c760d4f36ed6916e9eb997)

This PR adds the work-in-progress version of the pipeline workspace. I have made quite a lot of changes to the original code and there are still more needed, so I think it is better to post this now so that the remaining changes can be correctly reviewed. There is some commented code here that will be addressed soon.

(I reorganized the original code into subfolders including one named "class" and another "instance". This is due to the way the pipeline system is designed, because from a conceptual point of view the components that affect the behavior of the pipeline can be viewed as forming a class, and the components that affect the state of the pipeline can be viewed as forming an instance)

**The pipeline workspace is not yet functional. Do not load previous pipelines or they will be corrupted.**

### DIFF
--- a/src/app/medInria/CMakeLists.txt
+++ b/src/app/medInria/CMakeLists.txt
@@ -151,6 +151,8 @@ if (USE_PYTHON)
     target_link_libraries(${TARGET_NAME}
         medPythonBase
         medPythonTools
+        medPipeline
+        medPipelineBindings
         )
 endif()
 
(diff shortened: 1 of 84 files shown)


mathildemerle:
This PR is compiling on Ubuntu 20

PR #970, opened 2022-02-23
1st step to build with python on windows 

Florent2305:
change add_external_resources.cmake to reduce size of post-build script of medPythonBase on windows.

Add    **MEDCORELEGACY_EXPORT** macro on medExternalResources.h
Remove **MEDPYTHON_EXPORT** macro on some templates and macros
Add    **MEDPYTHON_EXPORT** on some functions and methods
Change **medPython_EXPORTS** to **medPythonBase_EXPORTS** in medPythonExport.h
Add target_link_directories for win32 in embed_python.cmake to find the python static lib stub for python dll implicit load.
Many changes in python_files_info_win.cmake but changes are not already used for the moment. changes are related to the line above.
Change template substitution type order in .i for a good  compilation.

### DIFF
--- a/src/layers/legacy/medCoreLegacy/medExternalResources.h
+++ b/src/layers/legacy/medCoreLegacy/medExternalResources.h
@@ -23,6 +23,7 @@
 
 
 #include <QString>
+#include <medCoreLegacyExport.h>
 
 namespace med
 {
@@ -30,12 +31,12 @@ namespace med
 // Returns the directory containing the external resources for the project or
 // one of its libraries. A null string is returned if the directory does not
 // exist.
-QString getExternalResourcesDirectory(QString libraryName = QString());
+    MEDCORELEGACY_EXPORT QString getExternalResourcesDirectory(QString libraryName = QString());
 
 // Returns the path of an external resource that was added using
 // add_external_resources in the cmake project. If the resource was added
 // through a library target then the library name must be provided. A null
 // string is returned if the resource is not found.
-QString getExternalResourcePath(QString filename, QString libraryName = QString());
+    MEDCORELEGACY_EXPORT QString getExternalResourcePath(QString filename, QString libraryName = QString());
 
 } // namespace med
--- a/src/layers/medPython/base/objects/medPythonModule.cpp
+++ b/src/layers/medPython/base/objects/medPythonModule.cpp
@@ -72,7 +72,14 @@ PyObject* Module::getModuleObject(QString name)
 
     if (!moduleObject)
     {
-        moduleObject = coreFunction(PyImport_ImportModuleLevel, qUtf8Printable(name), nullptr, nullptr, *list(), 0);
+        try
+        {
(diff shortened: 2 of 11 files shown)


Florent2305:
This PR **will need** to add **static stub .lib** to the **embedded python zip**  file

src/layers/medPython/base/objects/medPythonModule.cpp
```diff
@@ -72,7 +72,14 @@ PyObject* Module::getModuleObject(QString name)
...
     if (!moduleObject)
     {
-        moduleObject = coreFunction(PyImport_ImportModuleLevel, qUtf8Printable(name), nullptr, nullptr, *list(), 0);
+        try
+        {
+            moduleObject = coreFunction(PyImport_ImportModuleLevel, qUtf8Printable(name), nullptr, nullptr, *list(), 0);
+        }
+        catch (med::python::BaseException &e)
+        {
+            qDebug() << e.what();
```
fcollot, line 81: I don't think the exception should be stopped inside this function because it is low-level. More generally alll code using the medPython library should be aware that the API may raise exceptions, and so it is the calling code that should make sure to catch them.
(I know we had a crash relating to an exception that wasn't handled during initialisation and I'll fix that)

src/layers/legacy/medCoreLegacy/medExternalResources.h
```diff
@@ -23,19 +23,20 @@
...
+#include <medCoreLegacyExport.h>
 
 namespace med
 {
 
 // Returns the directory containing the external resources for the project or
 // one of its libraries. A null string is returned if the directory does not
 // exist.
-QString getExternalResourcesDirectory(QString libraryName = QString());
+    MEDCORELEGACY_EXPORT QString getExternalResourcesDirectory(QString libraryName = QString());
```
fcollot, line 34: Throughout medPython I did not indent namespace content (as I follow the google style guide). You indented these lines, but is it a typo or is that the style you follow?

fcollot (changes requested):

src/layers/legacy/medCoreLegacy/medExternalResources.h
Florent2305: Ho sorry it's my style code, but in your code I follow your style.
I will change it.

src/layers/medPython/base/objects/medPythonModule.cpp
Florent2305: ok I will remove it

fcollot (approved):
thanks for the changes and making this compile on windows

src/layers/legacy/medCoreLegacy/medExternalResources.h
Florent2305: done

src/layers/medPython/base/objects/medPythonModule.cpp
Florent2305: done

PR #971, opened 2022-03-01
[Paint] device ratio for cursor size

mathildemerle:
Based on PR:
 * https://github.com/Inria-Asclepios/medInria-public/pull/647
 * https://github.com/Inria-Asclepios/medInria-public/pull/730

Compute the right cursor size in the Paint tool, taking into account some screens (for instance Retina) which have different pixel ratio.

:m:


Florent2305 (approved):
ok

PR #976, opened 2022-03-04
[medPythonBaseException] fix exception message bug

fcollot:
The `qUtf8Printable` macro is the recommended way to convert a `QString` to a UTF8-encoded `const char` pointer (it is equivalent to calling `str.toUtf8().constData()`). The problem is the `toUtf8()` call creates a temporary `QByteArray`, and therefore the macro is not suitable for return values (as is done in `medPythonBaseException::what()`) because the pointer will not be valid after. To fix this problem we store the `QByteArray` in the exception instance instead of the initial `QString`.


Florent2305 (approved):
Good one

Issue #949, opened 2021-11-23
[Browser] ghost toolboxes from PACS/CD tab in DB and FS tabs


mathildemerle:
How to create the bug:
 * Open medInria 3.2 (built from scratch on my side)
 * Go to Browser workspace
 * Some tools ("Dicom servers settings" and "Series preview and import") from the PACS/CD tabs can be seen in the Database and File system tabs
 * If i switch to the PACS/CD tab and go back to the Database or File system tabs, the tools disappear

Example of the "ghost" toolboxes:
![Capture d’écran du 2021-11-23 14-02-36](https://user-images.githubusercontent.com/3910352/143029343-7b0042cd-e97d-41af-9f31-1d1c214d3a7e.png)

![Capture d’écran du 2021-11-23 14-03-24](https://user-images.githubusercontent.com/3910352/143029357-64497cc8-00a1-4525-8b34-efc668344003.png)

And the last tab, from which the toolboxes originally belong:
![Capture d’écran du 2021-11-23 14-03-40](https://user-images.githubusercontent.com/3910352/143029418-9d5383e3-2440-4916-9d2c-cc40f062a50d.png)

PR #988, opened 2022-03-23
[Browser] ghost toolboxes from PACS/CD tab in DB and FS tabs-master

mathildemerle:
Same as https://github.com/medInria/medInria-public/pull/987

This Pull Request solves 2 problems:
 * there were ghost toolboxes in tabs in Browser area (fix https://github.com/medInria/medInria-public/issues/949) -> no more ghosts
 * the default tab was defined before adding the data sources, which was not working, and users are often lost with Browser -> File system is the default one

:m:


Florent2305 (approved):
ok

PR #989, opened 2022-03-23
[Selector] nothing done if same tlbx switched

mathildemerle:
Following a comment by @fcollot in https://github.com/medInria/medInria-public/pull/967

Simplify code.

:m:


fcollot:
This isn't quite what I was saying. I was suggesting a way to avoid having to store the identifier (which is still done here).

mathildemerle:
@fcollot i used the suggestion you gave, and there are a lot of ways to do the same thing here... This solution is easy to understand and read, so it could be kept i think. 
Perhaps you could do your own PR if you want to enhance these kind of code?

fcollot:
The main idea of my suggested code (which isn't the same as here) was to not store the identifier because there is no need to do so. It wasn't about being more readable, it was to avoid having the selector toolbox store the identifier of a toolbox when it already has a pointer to that toolbox to retrieve the identifier.

Florent2305 (approved):
Ok

PR #991, opened 2022-03-25
[medPython] improve initialization

fcollot:
(based on https://github.com/medInria/medInria-public/pull/990)

The possibility to structure external resources using subdirectories makes the handling of the embedded Python much simpler. Also, some fixes were made to avoid crashes.

_**These changes require modified versions of the embedded Python that is downloaded. I'll notify when these are available.**_

Changes:

- If Python initialisation fails, do not run other Python related code (plugins, console etc.).

- Use the new external resources feature that allows subdirectories. This avoids the hack with the encodings package and greatly simplifies how we configure Python at initialisation.

- Convert module paths to native delimiters (fix for Windows).

- No need to list every single file of the embedded Python in CMake. The dependencies are simplified.


fcollot:
(the commit is https://github.com/medInria/medInria-public/pull/991/commits/078dc3e174ee311fabcbb2c299efddd95f977f6a)

PR #994, opened 2022-04-08
Resource libraries

fcollot:
(based on https://github.com/medInria/medInria-public/pull/993)
(commit (https://github.com/medInria/medInria-public/commit/51a7f1d918396abb65a8708d54c697d14d566583)]

Adds the concept of resource libraries which are libraries that must be packaged with the application but are not dependencies of the executable. The Python binding system creates shared libraries that are loaded by Python when the corresponding module is loaded. These libraries are not linked into the executable and therefore do not appear as dependencies. This causes a problem on macOS with dtkDeploy because it uses otool to find which libraries should be copied to the bundle. Plugins are handled differently: dtkDeploy imports all the plugins found in the designated directories. We can therefore use this to allow dtkDeploy to import resource libraries (and their dependencies) and then manually move them to the correct place in the bundle.

Notes:
  - changed the doc at the beginning of mac_packager.sh which was already outdated (the script works on directories only, not file paths)
  - the `set_lib_install_rules` macro is now a function and is properly indented.

### DIFF
--- a/packaging/apple/ApplePackaging.cmake
+++ b/packaging/apple/ApplePackaging.cmake
@@ -40,6 +40,7 @@ configure_file(
   
 configure_file(${CMAKE_SOURCE_DIR}/packaging/apple/mac_packager.sh.in 
   ${PROJECT_BINARY_DIR}/packaging/apple/mac_packager.sh
+  @ONLY
   )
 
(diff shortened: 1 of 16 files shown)


Florent2305 (approved):
Ok

PR #997, opened 2022-04-11
[medUtilitiesITK] binarize mask to 0/1

mathildemerle:
A lot of filters/toolbox can crash if the masks used are not 0/1 masks, as -1024/10000 or 0/255.

This PR adds a medUtilitiesITK fct to switch these masks to 0/1 if needed.

It will be needed on MUSICardio side in incomping PRs, and also in medInria.

:m:


fcollot:
statsROI is a tool to calculate values on images (volumes, averages etc.), not to generate new images. The function you added should be implemented as a separate filter.

mathildemerle:
I added separate files avoiding the use of statsROI

PR #1000, opened 2022-04-13
[medPython] more refactoring

fcollot:
(based on https://github.com/medInria/medInria-public/pull/996 and https://github.com/medInria/medInria-public/pull/999)

This is the second batch of changes in medPython. The first was the core, now it is mostly the layer above.

The main changes are:
- Remove the qt and medInria bindings from `medPythonTools` to make them separate modules.
- Make `medPythonTools` a proper Python module (the initialization code is moved from `medPythonTools.cpp` to the package's Python code (i.e. `__init__.py`))
- Refactor the cmake functions to allow `add_python_module` to handle single modules, packages, bindings, packages with bindings etc.
- Customize Python's import mechanism to allow loading dylibs on macOS.
- Handle the case of null values when converting `QVariant` to and from Python (bug found while working).
- Handle case of private Qt signals (necessary to connect to `aboutToQuit` from Python).
- No need to list all the embedded Python files as dependencies are now target-based instead of file-based.


mathildemerle:
is this needed to be reviewed, with #999?

fcollot:
https://github.com/medInria/medInria-public/pull/996, https://github.com/medInria/medInria-public/pull/999 and this PR need to be reviewed

src/layers/medPython/cmake/add_python_module.cmake
```diff
@@ -0,0 +1,169 @@
...
+# See LICENSE.txt for details in the root of the sources or:
+# https://github.com/medInria/medInria-public/blob/master/LICENSE.txt
+#
+# This software is distributed WITHOUT ANY WARRANTY; without even
+# the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+# PURPOSE.
+#
+################################################################################
+
+function(add_python_module target_name)
```
mathildemerle, line 16: it could be interesting to add explanations in comments in this file for future readers

src/layers/medPython/tools/python/console.py
```diff
@@ -14,6 +20,20 @@
...
+
+def initialize():
+    instance = Console(CONSOLE_TITLE, (CONSOLE_WIDTH, CONSOLE_HEIGHT))
+    instance.setShortcut(f'{CONSOLE_SHORTCUT}')
+    mainWindow = qt.qApp().getProperty('MainWindow')
+    mainWindow.connect('destroyed', lambda _ : instance.deleteLater())
+    instance.run()
+    medInria.logInfo(f'The Python console can be accessed with {CONSOLE_SHORTCUT}')
+    if failedPlugins:
+        print('** Some medInria plugins failed to load. Type "pluginsInfo()" for details. **\n\n')
```
mathildemerle, line 34: could you write "Some plugins failed" to help with rebranding/customization?

src/layers/medPython/tools/python/customImport.py
```diff
@@ -0,0 +1,9 @@
+import sys
+from importlib.machinery import *
+
+def installCustomImportHook():
```
mathildemerle, line 4: is it possible to add a comment to explain what that method does?

fcollot:
New changes:
- Remove unused compile definition.
- Add doc for `add_python_modules` command.
- Add doc for `installCustomImportHook` function.
- Add `.pyd` suffix to the custom import hook (problem found when testing on Windows).
- Modify a message in the Python console.

mathildemerle (approved):

mathildemerle:
Is this PR compiling? Do you need me to compile it from scratch on ubuntu for instance?

fcollot:
This PR compiles and runs but I have a new batch of changes so it may be preferable to wait for those. The new packages for the embedded Python are not available online yet so it only compiles if one has the package already in the build directory.

PR #1001, opened 2022-04-28
Improve architecture medSourcesSettings

Florent2305:
- create sub widget medSourcesSettingsHandlerWidget 
 - change behavior for medSourcesLoader about path of directory that contains sources descriptions.

### DIFF
--- a/src/layers/medCore/dataSource/medSourcesLoader.cpp
+++ b/src/layers/medCore/dataSource/medSourcesLoader.cpp
@@ -221,12 +221,32 @@ bool medSourcesLoader::setDefaultWorkingSource(QString const &instanceId)
 
 bool medSourcesLoader::setPath(QString path)
 {
-    bool bRes = false;
+    bool bRes = true;
 
-    if (QFile::exists(path + '/' + MED_DATASOURCES_FILENAME))
+    if (QDir(path).exists())
+    {
+        if (QFile::exists(path + '/' + MED_DATASOURCES_FILENAME))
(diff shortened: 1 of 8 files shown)


src/layers/medCore/dataSource/medSourcesLoader.cpp
```diff
@@ -221,20 +221,40 @@ bool medSourcesLoader::setDefaultWorkingSource(QString const &instanceId)
 
 bool medSourcesLoader::setPath(QString path)
 {
-    bool bRes = false;
+    bool bRes = true;
 
-    if (QFile::exists(path + '/' + MED_DATASOURCES_FILENAME))
+    if (QDir(path).exists())
+    {
+        if (QFile::exists(path + '/' + MED_DATASOURCES_FILENAME))
```
mathildemerle, line 228: maybe `QDir::separator()` instead of "/" ?

Florent2305: Thanks for this review. Here we handle Qt path format and not system path format then used QDir::separator() here will create a call, with both type of separator (on windows platform) , the QFile::exists. It's not recommended.
The design adopted here, in this class, is to deal with Qt path format internally and often it's possible and just adapte the path with [QDir::toNativeSeparators](https://doc.qt.io/qt-5/qdir.html#toNativeSeparators) when the software dialog with non Qt access file or to display it on widget, for example in medSourcesSettings.

PR #1002, opened 2022-05-02
Plugin aphp filter parameters

juliencastelneau:
**[WARN] - Work Only with qtdcm on branch PluginAPI**

- Add Parameters (with medAbstractParam mechanism) in Browser Area for Plugin APHP
- Implementation of methods  getMinimalEntries & getMandatoryAttributes by taking into account  the filters passed by users

TODO : Test with refresh button in GUI


Florent2305 (approved):
Ok because it's tested
Some bugs exists but will be removed by next developments

PR #1005, opened 2022-05-11
[medDoubleParameterL] fix wrong 0/0 values if attribute has only 1 value in a mesh

mathildemerle:
Same as this PR https://github.com/medInria/medInria-public/pull/886

If a mesh has an attribute with only one value (for instance 42), it was show as if the value was 0 (or one of the value already updated from an attribute with multiple values), and was not updated.

:m:


Florent2305 (approved):
ok

PR #902, opened 2021-09-02
[vtkImageView3D] Improvements

mjuhoor:
- Duplicate of https://github.com/Inria-Asclepios/medInria-public/pull/714

PR #1006, opened 2022-05-12
[Attributes] some mesh fields not displayed in color in 3D

mathildemerle:
Same as https://github.com/Inria-Asclepios/medInria-public/pull/736

Since this PR https://github.com/medInria/medInria-public/pull/902 some attributes of meshes were not displayed in color in 3D orientation. It was because vtkIdFilter vas not properly initialized. This PR solves this problem.

:m:


Florent2305 (approved):
Ok

Issue #1008, opened 2022-05-12
Medinria doest not read correctly dicom anymore


paulineMig:
Try to open a dicom instead of a volume you will have multiple files (one for each slice) this is due to the fact that in medAbstractDatabaseImporter.cpp the generateUniqueVolumeId function generate an unique key tintende to be unique for each volume however this key is create as follow: 
QString key = patientName+studyDicomId+seriesDescription+orientation+sequenceName+sliceThickness+rows+columns;

Which is problematic for anonymized dicom images because the name of the file is used in serieDescription which means a different key for each slice (dicom file name increment a number between each file of a volume). 
We should use another parameter than seriesDescription but which one is the best? SeriesNumber should do the job fine

PR #1009, opened 2022-05-13
[DataBaseImporter]: do not use same parameters to generate volume id of the medData object depending on the file extension. Dicom need a different approach.

paulineMig:
fix https://github.com/medInria/medInria-public/issues/1008
As explained in the issue as for now medinria is not able to read Dicom correctly.
Propose fix: keep the same key id for all extension except dicom one. in this case replace serieDescription by serieNumber + echoTime + flipAngle

### DIFF
--- a/src/layers/legacy/medCoreLegacy/database/medAbstractDatabaseImporter.cpp
+++ b/src/layers/legacy/medCoreLegacy/database/medAbstractDatabaseImporter.cpp
@@ -238,7 +238,8 @@ void medAbstractDatabaseImporter::importFile ( void )
 
             // 2.3) Generate an unique id for each volume
             // all images of the same volume should share the same id
-            QString volumeId = generateUniqueVolumeId ( medData );
+            bool isdicom = isDicomName(fileInfo.fileName());
+            QString volumeId = generateUniqueVolumeId ( medData, isdicom );
 
             // check whether the image belongs to a new volume
             if ( !volumeUniqueIdToVolumeNumber.contains ( volumeId ) )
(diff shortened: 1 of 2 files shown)


juliencastelneau (approved):
Ok for me
I introduced the bug few months ago when I modified the method `populateMissingMetadata`

src/layers/legacy/medCoreLegacy/database/medAbstractDatabaseImporter.cpp
```diff
@@ -1000,13 +1005,39 @@ QString medAbstractDatabaseImporter::generateUniqueVolumeId ( const medAbstractD
...
     return key;
 }
+
+//-----------------------------------------------------------------------------------------------------------
+/**
+* Look for dicom extension in file name
+* @param fileName - name of the file that we try to import
+* @return  true if dicom file, false otherwise
+**/
+bool medAbstractDatabaseImporter::isDicomName(const QString & fileName)
```
Florent2305, line 1035: Please replace by
```C++
bool medAbstractDatabaseImporter::isDicomName(const QString & fileName)
{
    QFileInfo info(fileName);
    return info.suffix().toLower() == "dcm";
}

Florent2305 (changes requested):
Thank you for this contribution.
Can you take into account my small change.

src/layers/legacy/medCoreLegacy/database/medAbstractDatabaseImporter.cpp
paulineMig: Done and good catch I do not know why i didn't do it like that from the start

Florent2305: Thanks

Florent2305 (approved):
ok

PR #1013, opened 2022-06-16
Insert some useful metadatas into the database

paulineMig:
-When writing a dicom file study date, study time, series times and series dates are needed if they are not provided by the data the writer uses the current time but the fractional part of a second could change between slices which can be problematic for some software like lead-DBS used by RebrAIn. Indeed for lead-DBS slices are not stacked if these values change.

-the kvp metadata information was not kept inside the database so the dicom writer was adding 1 by default to all ct dicom 

-Morever we now read patient age from the database so this information is accessible from the medinria metadata list interface

### DIFF
--- a/src/layers/legacy/medCoreLegacy/database/medLocalDbController.cpp
+++ b/src/layers/legacy/medCoreLegacy/database/medLocalDbController.cpp
@@ -107,16 +107,30 @@ bool medLocalDbController::createStudyTable()
 {
     QSqlQuery query(m_database);
 
-    return query.prepare(
-               "CREATE TABLE IF NOT EXISTS study ("
-               " id        INTEGER      PRIMARY KEY,"
-               " patient   INTEGER," // FOREIGN KEY
-               " name         TEXT,"
-               " uid          TEXT,"
-               " thumbnail    TEXT,"
-               " studyId      TEXT"
-               ");") &&
+    query.prepare(
+        "CREATE TABLE IF NOT EXISTS study ("
+        " id        INTEGER      PRIMARY KEY,"
+        " patient   INTEGER," // FOREIGN KEY
+        " name         TEXT,"
+        " uid          TEXT,"
+        " thumbnail    TEXT,"
+        " studyId      TEXT,"
+        " time         TEXT,"
+        " date         TEXT"
+        ");") &&
            execQuery(query, __FILE__, __LINE__);
+
+    // Get all the information about the table columns
+    query.prepare("PRAGMA table_info(study)");
+    if (!execQuery(query, __FILE__, __LINE__))
+    {
+        qDebug() << DTK_COLOR_FG_RED << query.lastError() << DTK_NO_COLOR;
+    }
+
+    this->addTextColumnToStudyTableIfNeeded(query, "time");
+    this->addTextColumnToStudyTableIfNeeded(query, "date");
+
+    return true;
 }
 
 bool medLocalDbController::createSeriesTable()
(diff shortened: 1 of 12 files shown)


src/layers/legacy/medCoreLegacy/database/medLocalDbController.cpp
```diff
@@ -107,16 +107,30 @@ bool medLocalDbController::createStudyTable()
...
+        " id        INTEGER      PRIMARY KEY,"
+        " patient   INTEGER," // FOREIGN KEY
+        " name         TEXT,"
+        " uid          TEXT,"
+        " thumbnail    TEXT,"
+        " studyId      TEXT,"
+        " time         TEXT,"
+        " date         TEXT"
+        ");") &&
            execQuery(query, __FILE__, __LINE__);
```
mathildemerle, line 121: it's also done line 125, is it normal?

src/layers/legacy/medCoreLegacy/database/medLocalDbController.cpp
```diff
@@ -107,16 +107,30 @@ bool medLocalDbController::createStudyTable()
...
+    query.prepare("PRAGMA table_info(study)");
+    if (!execQuery(query, __FILE__, __LINE__))
+    {
+        qDebug() << DTK_COLOR_FG_RED << query.lastError() << DTK_NO_COLOR;
+    }
+
+    this->addTextColumnToStudyTableIfNeeded(query, "time");
+    this->addTextColumnToStudyTableIfNeeded(query, "date");
+
+    return true;
```
mathildemerle, line 133: the function never return an error whereas it was possible before. Is it the behavior you were searching to create?

src/plugins/legacy/itkDataImage/writers/itkDicomDataImageWriter.cpp
```diff
@@ -107,72 +107,77 @@ QString itkDicomDataImageWriter::sopClassUID(QString modality)
 
 void itkDicomDataImageWriter::fillDictionaryFromMetaDataKey(itk::MetaDataDictionary &dictionary, bool &studyUIDExistance)
 {
+    bool studyTimeExistance = false;
```
mathildemerle, line 110: *existing

mathildemerle (changes requested):
I am testing this PR, i've got some errors trying to open in a view an old DICOM i just imported from Browser with this PR. I'm going to continue my tests, but this PR needs to be compatible with old DICOMs

src/layers/legacy/medCoreLegacy/database/medLocalDbController.cpp
paulineMig: I want the same comportement as createSeriesTable() so I copied the code from that function

src/layers/legacy/medCoreLegacy/database/medLocalDbController.cpp
paulineMig: Same as previous comment

paulineMig:
> I am testing this PR, i've got some errors trying to open in a view an old DICOM i just imported from Browser with this PR. I'm going to continue my tests, but this PR needs to be compatible with old DICOMs

can you send me the errors beacuse it seems to be working for me

src/layers/legacy/medCoreLegacy/database/medLocalDbController.cpp
paulineMig: I want the same comportement as createSeriesTable() so I copied the code from that function

src/layers/legacy/medCoreLegacy/database/medLocalDbController.cpp
paulineMig: same as previous comment

mathildemerle:
@paulineMig i had a problem with my build, it's ok now, i can import et read, no problem on my side

mathildemerle (approved):

PR #1014, opened 2022-06-22
[CentralWidget] merge recreate default widget, and switch after data removal

mathildemerle:
Apply https://github.com/medInria/medInria-public/pull/1012 which allow to recreate the central widget of a view (text + data open buttons). For instance in a toolbox if the inner widget of a toolbox has been changed, and we switch to an other toolbox.

On previous work https://github.com/medInria/medInria-public/pull/917 on master branch to solve some reinitialisation bugs of views with changed central widget (Filtering/Registration workspaces for instance).

OUTPUT view in Filtering, and FUSE view in Registration do not allow data drop, so i removed the opening buttons. OUTPUT view in Filtering in centered.

:m:


Florent2305 (approved):
Ok

PR #1015, opened 2022-06-22
[medUtilitiesITK] binarize mask to 0/1 --master

mathildemerle:
Adapt this PR to master branch: https://github.com/medInria/medInria-public/pull/997

> A lot of filters/toolbox can crash if the masks used are not 0/1 masks, as -1024/10000 or 0/255.
> 
> This PR adds a medUtilitiesITK fct to switch these masks to 0/1 if needed.
> 
> It will be needed on MUSICardio side in incomping PRs, and also in medInria.

:m:


Florent2305 (approved):
Ok

PR #1019, opened 2022-06-28
Fix mapWorldToDisplayCoordinates function 

paulineMig:
Calling mapWorldToDisplayCoordinates from medVtkView class leads to a crash. This is due to the fact that currentview is always null. To fix it we use imageView2D as done for mapDisplayToWorldCoordinates function

### DIFF
--- a/src/plugins/legacy/medVtkView/medVtkView.cpp
+++ b/src/plugins/legacy/medVtkView/medVtkView.cpp
@@ -257,7 +257,9 @@ QPointF medVtkView::mapWorldToDisplayCoordinates(const QVector3D & worldVec)
     // The following code is implemented without calling ren->SetWorldPoint,
     // because that generates an unnecessary modified event.
 
-    vtkRenderer * ren = d->currentView->GetRenderer();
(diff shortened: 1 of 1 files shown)


src/plugins/legacy/medVtkView/medVtkView.cpp
```diff
@@ -257,7 +257,9 @@ QPointF medVtkView::mapWorldToDisplayCoordinates(const QVector3D & worldVec)
     // The following code is implemented without calling ren->SetWorldPoint,
     // because that generates an unnecessary modified event.
 
-    vtkRenderer * ren = d->currentView->GetRenderer();
+    vtkRenderer * ren = d->view2d->GetRenderer();
+    if ( !ren )
```
mathildemerle, line 261: can you add brackets to these two blocks?

paulineMig: no problem :-)
I did not add thzm to stay consistent with the other if/return in the functions do you want me to add brackets around them too?

mathildemerle (approved):

PR #1021, opened 2022-07-05
Patient filter sq lite db

juliencastelneau:
1. Dev Filters in Legacy SQLite DB source (from APHP model) to enhance debugging of filters
2. Fix crash when a filtering is applied in browser while data is already in view (want to keep it)
[WARNING] - We are not totally sure that the bug is totally fixed. The code is more robust.


Florent2305 (approved):
ok

PR #1025, opened 2022-09-09
Remove Python from core (will be added as external project)

fcollot:
Python will be handled as a separate project from medInria.
(more info on this in the next PR)


mathildemerle (approved):

Florent2305 (approved):
I approve to move python to external project, then I approve this PR.

PR #1026, opened 2022-09-09
[Fedora] packaging libs in lib & lib64

mathildemerle:
I compiled and packaged medInria on Fedora 33.
Here is a line in the unix launcher that is needed to run medInria on Fedora. Libs in the package are in `lib64` and in `lib`.

:m:


Florent2305 (approved):
Ok

PR #1029, opened 2022-09-15
[Install rules]: fix link problems between medInria and external plugins

paulineMig:
Fix issues of type medCoreLegacy_LIBRARY-NOTFOUND when compiling with external plugins


mathildemerle (approved):
Tested, it solves the error on Ubuntu

Florent2305 (approved):
Ok

PR #1030, opened 2022-09-16
[iterativeClosestPoint] rewrite ICP code, reorganize toolbox, add fiducial registration error

mathildemerle:
From an old PR on MUSICardio that we did not duplicated to medInria https://github.com/Inria-Asclepios/medInria-public/pull/716

> To add fiducial registration error to ICP algorithm, it was necessary to recode the method.
> The toolbox was reorganized to have a more relevant disposition for the input parameters.
> The input parameters will be write in the transformation matrix file.
> The fiducial registration error is written in transformation matrix file and in the toolbox after process.

:m:

PR #1031, opened 2022-09-16
[iterativeClosestPoint] rewrite ICP code, reorganize toolbox, add fiducial registration error

mathildemerle:
Same PR as https://github.com/medInria/medInria-public/pull/1030 on medInria4.

> To add fiducial registration error to ICP algorithm, it was necessary to recode the method.
> The toolbox was reorganized to have a more relevant disposition for the input parameters.
> The input parameters will be write in the transformation matrix file.
> The fiducial registration error is written in transformation matrix file and in the toolbox after process.

:m:


Florent2305 (approved):
ok

Issue #1027, opened 2022-09-12
[Import] cant import mha with accented path  


mathildemerle:
Tested on Fedora 33, i tried to import a mha and i got:
`Reader was unable to read at least:  \"/home/mathildemerle/Téléchargements/Belzebuth/pact (polygon roi).mha\"`

I moved the data to a non accented location and no more import failure.
Also, this does not happen with vtk import.

PR #1032, opened 2022-09-19
[Path] switch to utf8 to accept accented paths

mathildemerle:
Fix https://github.com/medInria/medInria-public/issues/1027

Solve import error when there are accents in a data path to import.

:m:


Florent2305 (approved):
OK

PR #1033, opened 2022-09-19
[Path] switch to utf8 to accept accented paths

mathildemerle:
Same as https://github.com/medInria/medInria-public/pull/1032 on master branch.

:m:


Florent2305 (approved):
Ok

PR #1035, opened 2022-09-22
[VTK] patch to compile with GCC 10 - fedora

mathildemerle:
On Fedora 33 with gcc10 (we are at gcc9 on Ubuntu 20 for instance), the gcc version number is not found... because it has two digits. This is because we are on VTK 8.1.2 which is a bit old. This will be solved in new versions.

:m:


Florent2305 (approved):
ok

Issue #793, opened 2020-11-10
[VariationalSegmentation] create a white half-box


mathildemerle:
[Test of the Ubuntu20 Jenkins binary on a Ubuntu20 computer]

 * Open a data in Variational Segmentation
 * Create a segmentation
 * Click on "Apply Segmentation" or "Save Mask" -> there are 2 white borders at bottom and right edges of the output mask.

Same error on medInria and MUSICardio.

mathildemerle:
Testing the current state of medInria/MSC, we still have this bug of white borders

PR #1041, opened 2022-10-06
[Variational] GUI, applied data saved, data name, spacing error

mathildemerle:
This PR:
 * fix https://github.com/medInria/medInria-public/issues/793 removing the white plans on output mask
 * enhance the GUI, the first button is the saving of the mask, before the application of the mask on the view, and exported data name shorten
 * the application of the mask save the applied data, not the mask, which is done with the mask button

:m:


fcollot (approved):

Florent2305 (approved):
ok

PR #1042, opened 2022-10-07
[Combobox] right size with long data array name

mathildemerle:
Solve toolbox width error in case of opening data with long data array name, before opening the toolbox.

:m:


fcollot (approved):

Florent2305 (approved):
Ok

PR #1047, opened 2022-10-21
Add PYNCPP external project

fcollot:
The PYNCPP project compiles it's own version of Python (or installs one on Windows), with ._pth configuration file support on all platforms, and provides an API to embed Python in a CPP application.
(This only adds the external project. The runtime initialisation of Python and the packaging aspects are not handled here)


Florent2305 (approved):
Ok cool

superbuild/projects_modules/medInria.cmake
```diff
@@ -93,6 +93,7 @@ set(cmake_cache_args
   -DTTK_DIR:PATH=${TTK_DIR}
   -DVTK_DIR:PATH=${VTK_DIR}
   -DQt5_DIR:PATH=${Qt5_DIR}
+  -DPYNCPP_DIR:PATH=${PYNCPP_DIR}
```
mathildemerle, line 96: no `if(${USE_Python})` as for FFmpeg?

juliencastelneau (approved):

PR #1049, opened 2022-10-24
[Meshes] Fix mesh readers -- master

mathildemerle:
Adapt this PR https://github.com/medInria/medInria-public/pull/954 to master branch.

> - In volume meshes, edges, triangles, ridges are read and not only tetrahedra as before.
> - Similarly in surface meshes, edges, ridges are read and not only triangles.
> - Add vtu reader and writer for mesh of type unstructured grid
> - Prevent use of vtp writer for unstructured grid and vtu writer for polydata
> - Check the version of the vtk file to avoid crash due to the new format introduced in vtk 9


Florent2305 (dismissed):
Ok

PR #1051, opened 2022-10-27
[medAbstractData]: add drivers capable of on offscreen rendering

paulineMig:
On virtual machine using mesa/x.org drivers each time a thumbnail was generated an orphaned widget was seen on the screen.
To fix this issue the driver is added to the list of drivers capable of on offscreen rendering in generateThumbnailInGuiThread method.


mathildemerle (approved):

Florent2305 (approved):
Ok

Issue #1046, opened 2022-10-21
[macOS] packaging error at opening: cannot load library, no plugins at all


mathildemerle:
At launch of the macOS binary, there are no plugins including those from medInria and our external project:

`WARN  - Wed Oct 19 14:20:44 2022 - "Unable to load /Applications/MUSICardio 2.app/Contents/PlugIns/libmedItkArithmeticOperationProcessPlugin.dylib - Cannot load library /Applications/MUSICardio 2.app/Contents/PlugIns/libmedItkArithmeticOperationProcessPlugin.dylib: (dlopen(/Applications/MUSICardio 2.app/Contents/PlugIns/libmedItkArithmeticOperationProcessPlugin.dylib, 0x0009): Library not loaded: /System/Volumes/Data/builds/workspace/music-macos-free/build/medInria-build/lib/resource_libs/libmedWidgets.dylib\n  Referenced from: /Applications/MUSICardio 2.app/Contents/PlugIns/libmedItkArithmeticOperationProcessPlugin.dylib\n  Reason: tried: '/System/Volumes/Data/builds/workspace/music-macos-free/build/medInria-build/lib/resource_libs/libmedWidgets.dylib' (no such file), '/usr/local/lib/libmedWidgets.dylib' (no such file), '/usr/lib/libmedWidgets.dylib' (no such file))" `

It could be linked to https://github.com/medInria/medInria-public/pull/1039 which have been removed from `mac_packager.sh.in`. I am trying to test on our jenkins macOS item, but it's not ideal.

For now, to compile (on zsh):

```zsh
resourceLibsDir=@medInria_BINARY_DIR@/lib/resource_libs #resources did not exist 
injectDirs="$injectDirs -inject-dir=$resourceLibsDir"
```

and (changing the app name)

```zsh
echo "##########################"
contentsDir=@medInria_BINARY_DIR@/bin/TmpInstall/MUSICardio.app/Contents
ls $contentsDir/PlugIns

echo "########################## 2 "
dylibNames=$(ls $contentsDir/PlugIns/*.dylib)
for lib in $dylibNames; do
    echo $lib
    libname=${lib##*/}
    installedName="@executable_path/../PlugIns/$libname"
    install_name_tool -id $installedName $contentsDir/PlugIns/$libname
done

# Resource libraries are sent as plugins to dtkDeploy (above) because they are
# not dependencies of the executable (dtkDeploy uses otool on the executable to
# find which libraires should be imported, so it will not find resource
[...]

PR #1053, opened 2022-11-09
fix issue with dropped resource libs support

fcollot:
Fixes https://github.com/medInria/medInria-public/issues/1046 (except for non-legacy plugins, which is a separate problem)


mathildemerle (approved):

Florent2305 (approved):
ok

PR #1054, opened 2022-11-10
Fix plugin paths on for developer builds

fcollot:
The medInria launcher creates two environment variables for plugin paths called `MEDINRIA_PLUGINS_DIR` and `MEDINRIA_PLUGINS_DIR_LEGACY`, but `medPluginManager` only uses the legacy one. The consequence of this is that non-legacy plugins were never loaded in developer builds (and as a consequence of that, the bug related to loading of non-legacy plugins was not noticed because it happens only in packaged builds).

It makes more sense to have one path variable anyways, so this PR merges them together.


fcollot:
I modified the title of this PR, as this bug was present on all platforms.

mathildemerle (approved):

fcollot:
Don't merge this yet, it seems I have missed something.

fcollot:
After discussion with @Florent2305 it seems this is only a minor bug. We handle both dtk and dtk2 plugins (legacy and non-legacy), and both path variables are handled differently. What this means is the dtk2 plugins are loaded separately (and successfully), but in the MacOS package all plugins are located in the same directory so the dtk plugin loader tries to load them and complains.

fcollot:
(I am unable to close this PR, github is not working right on my browser today)

fcollot:
Yes only that PR

Issue #491, opened 2019-11-19
Incorrect zoom when loading large images


ocommowi:
This bug is there since medinria 3.0. If you load a large image (let's say CT_mouse.mha in medinria-data), the zoom is incorrect. Until medinria 2.2.3, the zoom was such that the slice shown by default used the whole view, now it is not. It seems the zoom is set for all three planes to the largest field of view between sagittal, axial and coronal.

Here are two images showing before and after (2.2.3 vs 3.x)

<img width="898" alt="Capture d’écran 2019-11-19 à 12 04 11" src="https://user-images.githubusercontent.com/2983208/69142217-a2d6a600-0ac6-11ea-84e5-4dfbecfecd9a.png">
<img width="904" alt="Capture d’écran 2019-11-19 à 12 06 40" src="https://user-images.githubusercontent.com/2983208/69142223-a66a2d00-0ac6-11ea-813e-4aa3b6fd062e.png">

Florent2305:
move to 3.1.1

mathildemerle:
I was looking at this problem. Is there a medInria branch of 2.2.3? I'm reading the old code on https://github.com/medInria/medInria-public/tree/QT5_dtk1 but it's perhaps not the best one.

By the way, it could be interesting to tag each release branch. We have a wiki page with that on our side:

> // Create the release branch
> git fetch origin
> git checkout origin/branch
> git checkout -b X.Y.Z
> git push origin X.Y.Z
> 
> // If we want to save tags
> git tag -a vX.Y.Z -m "vX.Y.Z"
> // check tag if needed
> git tag          
> git push origin vX.Y.Z
> // If need to update a tag (bugfix)
> git tag -a -f vX.Y.Z de24b6ac65f2e23563c7fe691da0871e020eb24b # commit number

It will be easier to look at old code, and check if a bug was in a previous release, etc.

Florent2305:
After a meeting with @ocommowi  we have decided to downgrade issue level from major to minor and then to postpone to 4.0

nbaraka:
I would like to check this. Where can I load the data test :  **CT_mouse.mha in medinria-data**
Thanks

mathildemerle:
@nbaraka in https://github.com/medInria/medInria-data :)

PR #1064, opened 2022-12-06
[Zoom] proper zoom displaying data on xy and not xyz

mathildemerle:
Fix https://github.com/medInria/medInria-public/issues/491

It was a very old bug in medInria from 3.0. The zoom was incorrectly set at opening of data, specially some data as the medInria data test "CT_mouse.mha". 

In this PR, the zoom value is calculated through only x and y values, not z. 

Let me know if you think this new behavior could create any problem elsewhere.

:m:


Florent2305 (approved):
ok,

PR #1065, opened 2022-12-06
[Zoom] proper zoom displaying data on xy and not xyz --master

mathildemerle:
Same as https://github.com/medInria/medInria-public/pull/1064 on master.

It was a very old bug in medInria from 3.0. The zoom was incorrectly set at opening of data, specially some data as the medInria data test "CT_mouse.mha".  In this PR, the zoom value is calculated through only x and y values, not z.

:m:


Florent2305 (approved):
ok

Issue #644, opened 2020-04-29
[Quit] Run N4 bias -> quit medInria -> quit anyway -> does not want to 


mathildemerle:
If you run the N4 bias correction process for instance, which is long, and click on the exit medInria button, then click on "Quit", the application does not stop but freezes then display "medInria does not respond - force to quit".

Happens at least on 3.1.1.

PR #1069, opened 2022-12-08
[N4Bias] manage process cancel if the user wants to quit the app

mathildemerle:
Fix https://github.com/medInria/medInria-public/issues/644

We can't quit a medRunnableProcess/QRunnableProcess naturally, but some checkpoints can be added in the running process to quit when reached.

When the user tries to close medInria, `medMainWindow::closeEvent` is sent, then `medJobManagerL::dispatchGlobalCancelEvent` emits a cancel signal, which is caught by `medRunnableProcess::onCancel`.

At the end, in `medAbstractProcessLegacy`  processes, `onCanceled` are run if it exists. Since we can't close directly the process, we can add checkpoints in the running method to test after or before some long algo if the signal has not been caught. For instance here in `medN4BiasCorrection::update` i added some:

```c++
 if (d->cancelAsked)
 {
        return medAbstractProcessLegacy::FAILURE;
 }
```

PS : i removed the _p file which was not necessary, i did not changed the algo, i just added the cancel tests.

:m:


Florent2305 (approved):
Ok

Issue #674, opened 2020-05-06
Reslice : top Left View Display change after spacing change


nbaraka:
When a change is done in spacing, the first top left view has its size that is changed. This happen only for this view.

PR #1072, opened 2022-12-08
[Reslice] fix zoom error + code clarity --master

mathildemerle:
Same as https://github.com/medInria/medInria-public/pull/1071

Fix https://github.com/medInria/medInria-public/issues/674

When the user uses the dimension or spacing tool, there is no more a graphic zoom bug on the views.

To find the solution, i needed to clean the code (specially askedSpacingOrSizeChange) to create a clear separation between spacing change and size (previously 'dimension') change, because a lot of the widgets have been wrongly name only with Spacing, instead of both. This is surely due to code history, spacing was coded before.

:m:


Florent2305 (approved):
ok

Issue #726, opened 2020-06-04
[Paint] creation of mask data at Paint/Erase


mathildemerle:
From an idea of @Florent2305:

Create the mask data in Paint when the user clicks on "Paint/Erase", not at the first click on the image. The creation of the mask is too slow, and the drawing not smooth at the beginning.

PR #1074, opened 2022-12-09
[Paint] create mask before trying to paint + hide a button

mathildemerle:
Same as https://github.com/medInria/medInria-public/pull/1073

Fix https://github.com/medInria/medInria-public/issues/726

Create the mask data in Paint when the user clicks on "Paint/Erase", not at the first click on the image. The creation of the mask is too slow, and the drawing not smooth at the beginning.

Also, switching from Magic Wand to Paint/Erase was not hiding the removeSeed button.

:m:


Florent2305 (approved):
Ok

PR #1063, opened 2022-12-02
[Four-views] solve crash closing the views --3.3

mathildemerle:
Same as https://github.com/medInria/medInria-public/pull/1060

Fix https://github.com/medInria/medInria-public/issues/995

The views are already cleaned at closing through a signal set in `medViewParameterGroupL::addImpactedView`. 
Removing them manually in `~medViewParameterGroupL` leads to trying to remove already removed views, generating a crash.

:m:

PR #1075, opened 2022-12-12
[3.3.1] release notes & version number with four view crash-fix

mathildemerle:
Following this crash-fix : https://github.com/medInria/medInria-public/pull/1063

This PR allows to update the release notes and version number.

:m:


Florent2305 (commented):
ok

PR #1080, opened 2022-12-16
[Git diff] solve differences between musicardio to medInria4

mathildemerle:
Same as https://github.com/medInria/medInria-public/pull/1079

I did some `git diff` between MUSICardio on master branch (3.4) and medInria3.4 for forgot changes. This aims to avoid differences between both code and to avoid having merge problems with synchronisation.

:m:


Florent2305 (approved):
ok

PR #1082, opened 2022-12-19
Set cmake minimum version to 19 for medInria project

fcollot:
The minimum required cmake version is still set at 3.3 on medInria 3. This means that for any cmake policy introduced after 3.3 the default behaviour of the policy will be chosen unless it is manually set. This is causing issues with the python embedding code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@
 #
 ################################################################################
 
-cmake_minimum_required(VERSION 3.3)
+cmake_minimum_required(VERSION 3.19)
 
 if(POLICY CMP0020)
   cmake_policy(SET CMP0020 NEW)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -13,7 +13,7 @@
 #
 ################################################################################
 
-cmake_minimum_required(VERSION 3.3)
+cmake_minimum_required(VERSION 3.19)
 
 if(NOT DEFINED ${medInria_VERSION})
     set(medInria_VERSION 3.3.0)


mathildemerle:
On this PR (meaning: on MUSICardio) we are at 3.15: https://github.com/medInria/medInria-public/pull/1079
On our doc we are at 3.19, idem for medInria: https://github.com/medInria/medInria-public/wiki/A-Quick-Dev-Guide---prerequisites
So it would be interesting to know precisely. Do you know the real minimum?

fcollot:
First of all the medInria doc does not specify which branch it applies to. We currently have medInria 3 and 4 in development and it is normal to have different requirements for both, so we should maybe clarify what version that doc applies to.
The bug I had concerns cmake behaviour modified in 3.14 and for which the old behaviour is the default policy, so the minimum version for that bug only is 3.14. Nevertheless I successfully compiled with cmake 3.17 on Ubuntu and 3.20 on macOS, so I chose 3.17 as the minimum version because it is tested and sufficiently old to be a light requirement, but 3.19 would be even better if everyone agrees (for the medInria 3.4 branch).

mathildemerle:
Indeed, the documentations are for the current released versions: 3.3 medInria and MUSICardio. The 1079 PR updates the code for 3.4, and medInria4 can be changed independently. 
For now, every OS compile on for medInria and MSC are on or above CMake 3.19. Maybe we can stay on it, since we know it's ok?

fcollot:
I was already pushing for 3.19 last year so I have no problem with it, on the contrary. I'll make the change.

fcollot:
done

mathildemerle (approved):

PR #1083, opened 2022-12-19
[DCM] do not corrupt exported/imported dcm with odd axis

mathildemerle:
Exporting and importing DICOM from medInria3.4 i found a bug with some of my files with these kinds of error messages "DEBUG - ven. déc. 16 14:39:11 2022 - "itk::ERROR: DCMTKImageIO(0x7fcd340207f0): Ambiguous slice stack direction: -0.663234 0.582022 0.470501""

After some researches, it is linked to https://github.com/medInria/medInria-public/pull/1003

The solution is to continue the `GetPositionOnStackingAxisForImage` method, even if the axis is not what we may ask.

:m:

### DIFF
--- a/src/layers/legacy/medImageIO/itkDCMTKImageIO.cpp
+++ b/src/layers/legacy/medImageIO/itkDCMTKImageIO.cpp
@@ -673,7 +673,7 @@ void DCMTKImageIO::DetermineOrientation()
 double DCMTKImageIO::GetPositionOnStackingAxisForImage (int index)
 {
     // Get maximum absolute value, which is the closest to an axis
-    auto result = std::max_element(m_Direction[2].begin(), m_Direction[2].end(), [](int a, int b)
+    auto result = std::max_element(m_Direction[2].begin(), m_Direction[2].end(), [](double a, double b)
     {
         return std::abs(a) < std::abs(b);
     });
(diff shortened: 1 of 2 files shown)


src/layers/legacy/medImageIO/itkDCMTKImageIO.cpp
```diff
@@ -679,15 +678,6 @@ double DCMTKImageIO::GetPositionOnStackingAxisForImage (int index)
     closestAxis[0] = round(m_Direction[2][0]);
     closestAxis[1] = round(m_Direction[2][1]);
     closestAxis[2] = round(m_Direction[2][2]);
```
paulineMig, line 680: Maybe best to determine which direction is the closest to an axis by determining the max of m_Direction[2] instead of removing juste the if.

mathildemerle:
@paulineMig following your advice, i rewrote the function to get the absolute maximum value position, meaning: the actual value closest to an axis

mathildemerle:
I am not sure about the returning value however, i did not have mirrored data to test

mathildemerle:
@paulineMig thx for your comment. 
The first part (computation of principalAxis index) was already done in the PR. Concerning the second part, a part was missing, as seen in the old code for this function. I replaced that part with a modernized c++ way to convert the `string` metadata to a vector of `double`, and get back the index value.
Let me know if it's ok!

paulineMig:
@mathildemerle The first part of the code does not compute the principal axis index as expected
This is the result obtain with your method to compute principal axis index and mine on a problematic data 
Direction: -0.0315755 0.588033 0.80822
Principal axis: 0
Principal axis expected: 2
Position: -108.25556905 -43.47483433 160.40608552
Return value: 160.406
Return value expected: -108.256

mathildemerle:
Indeed, there was a bug if the max absolute value was not in first position in the vector (as in my data), thank you for that. I added a commit to debug that

paulineMig (approved):

Florent2305 (approved):
ok

PR #1084, opened 2022-12-19
[Dicom] Fixing mirrored DICOM images and their origin

mathildemerle:
From PRs https://github.com/medInria/medInria-public/pull/1003 and https://github.com/medInria/medInria-public/pull/1083

> The method that gets the position of a slice on the stacking axis was z-specific and only worked on axial acquisition. This method was involved in both the determination of the origin of the image and the ordering of slices in the buffer.
> 
> Furthermore, the aforementioned slice ordering code is redundant with the code above it, which led to a double order inversion instead of a simple inversion of the order in which slices are stacked in the buffer, causing the images to be mirrored in case the stacking direction is the opposite of the acquisition direction.

:m:

### DIFF
--- a/src/layers/legacy/medImageIO/itkDCMTKImageIO.cpp
+++ b/src/layers/legacy/medImageIO/itkDCMTKImageIO.cpp
@@ -670,7 +670,6 @@ void DCMTKImageIO::DetermineOrientation()
     }
 }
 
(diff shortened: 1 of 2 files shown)


Florent2305 (approved):
ok

PR #1085, opened 2022-12-21
[macOS] enable shortcuts

mathildemerle:
From an issue on https://github.com/Inria-Asclepios/medInria-public/issues/169

On macOS the shortcuts in Qt applications are disabled by default: https://doc.qt.io/qt-6/qshortcut.html#details

This PR adds the methods which enable shortcuts.

I put the PR in draft since it needs to be tested. It can be tested with the Exit pop-up.

:m:

### DIFF
--- a/src/app/medInria/main.cpp
+++ b/src/app/medInria/main.cpp
@@ -29,6 +29,10 @@
 #include <medSettingsManager.h>
 #include <medStorage.h>
 
+#ifdef Q_WS_MAC
+extern void qt_set_sequence_auto_mnemonic(bool b);
+#endif
+
 void forceShow(medMainWindow& mainwindow )
 {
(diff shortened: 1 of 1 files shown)


mathildemerle:
Tested on macOS: 
 * on master branch, le native macOS menu bar that we use have its own keyboard shortcut to access: `control + fn + F2`
 * as said by [Qt shortcut documentation](https://doc.qt.io/qt-6/qshortcut.html#details) on macOS even if we enable the shortcuts, the underlined letters as done on linux and windows are not shown in the native menu bar on macOS
 * this qt setting in this PR could be used by other widgets, for instance the "Do you really want to exit?" popup. However, it's not working...
 
So, the qt shortcut system on macOS is not stable for now on Qt5. We can keep this PR to test when we'll be on Qt6. I put this PR as invalid for now.

PR #1086, opened 2023-01-02
[Registration] selector toolbox empty connect

mathildemerle:
Avoid trying to connect with an empty toolbox in some cases.

:m:


Florent2305 (approved):
ok

PR #1087, opened 2023-01-02
[Registration] selector toolbox empty connect --master

mathildemerle:
Same as https://github.com/medInria/medInria-public/pull/1086 but on master branch.

Avoid trying to connect with an empty toolbox in some cases.

:m:


Florent2305 (approved):
ok

PR #1091, opened 2023-01-26
[Themes] introduce theme system and add grey and light themes

mathildemerle:
⭐ Some users prefers to work on a lighter theme in their applications, and modularity/rebranding is an objective in medInria 4.0. We developed few years ago a theme system in MUSICardio, that has not been added in medInria when i brought our enhancements/debugs here. This was due to a lack of an icon theme mechanism: we needed to manually code the switch between icons according to the theme in the code. Thanks to the PR 1090, this is solved, and unblock the problem. I personally work a lot on light themes, and @paulineMig too. These PR will allow us to switch to this enhanced theme system, and we'll less depend on changes in our forks.

:suspect: "Gneugneu i hate light theme" -> you're not used, but some people likes it, and you can stick to the dark one ;)

**Based on the PR:** 
 * https://github.com/medInria/medInria-public/pull/1090. 
 * This PR could be rebased with 1090.

**Commits:**
 * https://github.com/medInria/medInria-public/commit/7022bd04ec951c0f902371eb05a5147c3895b84d
 * https://github.com/medInria/medInria-public/pull/1091/commits/9a976dc7ae734a0bd7ad8412a62ab31c0e3bbb76

**This PR does:**
 * Enhance QMenuBar categories.
 * Debug the fullscreen button behavior if the application starts in fullscreen mode.
 * Add an `initializeThemes` method in `medApplication` which centralizes the switch between themes (stylesheets) and icon themes (previously introduced in PR 1090).
 * `medSplashScreen` and `medSliderL`: adapts colors to the current theme.
 * Finish the sorting of icons started in PR1090 in black or white icon themes.
 * Add 2 themes from MUSICardio: a grey one and a light one. The original medInria one is the dark one.
 * Adapt the 3 themes to the icon themes.
 * Debug an icon problem on windows when we change the icon.
 * Add a combobox widget in start-up settings to switch between themes, in `medStartupSettingsWidget`.
 
**About the need to restart to see the new theme:**
In `medStartupSettingsWidget::write()` we could call `medApplication::initializeThemes()` to update in real time the theme (like in Visual Studio Code for instance) but we do not have access to medApplication. We can pass `QApplication` with `qApp` parameter in `medHomepageArea::onShowAreaSettings()` (since medApplication is a QApplication) but i could not access to medApplication. So, the user needs to restart the application. I added a label in the settings popup to explain the need to restart.

**What are the next steps?** 
[...]

Issue #866, opened 2021-04-15
[dmg] modularity of dmg design


mathildemerle:
I was working on the dmg creation done through [mac_packager.sh.in](https://github.com/medInria/medInria-public/blob/master/packaging/apple/mac_packager.sh.in) because on the MUSICardio fork we still had the medInria logo and names on our dmg. This is an important topic for the modularity needed in medInria 4.x.

Firstly i tried to update the `.sparseimage.gz` found in `src/utils/osx_packaging` but this need to be graphically done with DropDMG for instance, on a mac, and there were some logo positions and icon errors. This is not really what i was looking for. This is not really modular.

### dmgbuild
Finally, i found the [dmgbuild](https://pypi.org/project/dmgbuild/) tool.

 * `pip3 install dmgbuild` -> this would be a small prerequisites for macOS, since we already have Python as pre-requisites (pip).
 * Create a `settings.json` file in `packaging/apple`. It can easily be adapted by forked-projects of medInria.
 * The volume should be mounted, with a name like `"medInria @MEDINRIA_SUPERBUILD_VERSION@"` as done in `mac_packager.sh.in`, before using dmgbuild.
 * In medInria, the command line would be `dmgbuild -s ../packaging/apple/settings.json "medInria @MEDINRIA_SUPERBUILD_VERSION@" medInria.dmg`.
 * For now, i'm using this dmgbuild line outside of medInria/MUSICardio, in our Jenkins item, but it could be added in `mac_packager.sh.in` file for instance.
 * The tool created a nice dmg with the right icons and names.
 * The tool can be useful also for medInria itself to easily update the dmg design.

Example of `settings.json` that i created for our fork (the height value needs to be "background height + 20 pixels" to take the window bar into account):
```json
{
    "title": "MUSICardio3",
    "background": "../src/app/medInria/resources/pixmaps/MUSICardio-logo-dmg.png",
    "icon": "../src/app/medInria/resources/pixmaps/MUSICardio_dmg_mountedimage.icns",
    "format": "UDZO",
    "compression-level": 9,
    "window": { "position": { "x": 100, "y": 100 },
                "size": { "width": 640, "height": 300 } },
    "contents": [
        { "x": 140, "y": 165, "type": "file", "path": "/Volumes/MUSICardio/MUSICardio.app" },
        { "x": 480, "y": 165, "type": "link", "path": "/Applications" }
    ]
}
```

### Suggestions
[...]

Issue #864, opened 2021-04-07
Objectives 4.x


Florent2305:
This issues lists more or less global objectives to achieve during later subversions of medInria 4.
### Scenes
  - [ ] extract scenes functions (toXMLNode, fromXMLNode, medAbstractLayeredView::write, medAbstractLayeredView::restoreState). 
  - [ ] parameters don't format data to xml directly. We must use new software architecture. Parameters must exchange abstract representation with a specific IO class to read/write files in XML or other future format.

### Browser
  - [ ] Allow to view network and removal drives (related to #814)

### Notifications
  - [ ] Add a system of notifications 
 
### Filtering
  - Remove isolated voxels/pixels
    - [ ] Add capability to support non-binary mask like (-1024 of background, 10000 positive value) (related to #798)
  - Mask application
    - [ ] Add capability to support non-binary mask like (-1024 of background, 10000 positive value) (related to #588)
 
### Copyright and Licences
  - [ ] Enhance copyright (related to #543)
  - [ ] List licences used by third-parties
  - [ ] Provide links to sources of third-parties on website

mathildemerle:
Added some suggestions of objectives: 

### Modularity 
 - [ ] [[dmg] modularity of dmg design](https://github.com/medInria/medInria-public/issues/866) to change the dmg icons, names.
 - [ ] Icons, names, description, theme needs to be easily changed in fork-projects. Idea: directory/json file with all the icons, text, image, etc.

### Packaging
 - [ ] Linux: a real package to install the software 

### GUI
 - [ ] [Improve Home page ergonomics and style](https://github.com/medInria/medInria-public/issues/446)

mathildemerle:
The plugin shop could be listed?

PR #1094, opened 2023-02-06
[Custom] configure setting file to customize medInria 4

mathildemerle:
The commit : https://github.com/medInria/medInria-public/commit/f05491a24d0757d85bb4ab3bd3fe3c9f9ff487f9
PR based on PR 1090 and https://github.com/medInria/medInria-public/pull/1091

One of the [goals of medInria4](https://github.com/medInria/medInria-public/issues/864) is an easy customization of the app for external projects using medInria. 

### The customization system
This PR adds a `settings.json` file at root of medInria, set for medInria default "style": name, description, icons, logos, release notes, etc. CMake reads this json file and extracts the information, passes them also to medInria c++ parts. These values allow to configure the application. 

I added a medInria version of the logo for light background/theme.

This settings file can be adapted as i did for [MUSICardio](https://github.com/LIRYC-IHU/msc_config/blob/main/settings.json). The devs of an external project of medInria, can create their own repo with dedicated info, to access the settings files easily. 

### For instance to compile med4 with a rebranding
 * git clone medInria4
 * enter medInria4, git clone customization repo
 * replace the settings file at root of medInria by the one from the customization directory
 * build the application

### What remains to be done
 * i tested on Ubuntu 20.04. I/we need to test on the other platforms
 * Specifically on macOS there is the dmg style problem which remains: https://github.com/medInria/medInria-public/issues/866. I added this system in MUSICardio but it needs to be tested and adapted to medInria configuration

:m:


Florent2305 (commented):
We must keep this code base but, I don't agree to merge it directly.
I would like to create a new PR based on this one, with changes concerning the CMake part.

PR #1096, opened 2023-02-08
[CMake] minimum version at 3.19 --master

mathildemerle:
Same as https://github.com/medInria/medInria-public/pull/1095

Switch cmake minimum to 3.19.

:m:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@
 #
 ################################################################################
 
-cmake_minimum_required(VERSION 3.3)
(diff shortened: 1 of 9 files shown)


fcollot:
It would be better to remove the lines `cmake_minimum_required(VERSION ${CMAKE_MINIMUM_REQUIRED_VERSION})` entirely as they serve no purpose now since they are simply repeating the same call as the top level one.

mathildemerle:
@fcollot done

PR #1102, opened 2023-02-09
[Custom] add possibility to have an expiration date to the app

mathildemerle:
Based on https://github.com/medInria/medInria-public/pull/1094

Commit: https://github.com/medInria/medInria-public/commit/4e22ca6e0cfa76de74bbac9112626e9b612f1db6

From a feature on MUSICardio, this PR adds the possibility (by default deactivated) to add an expiration time to a binary. If the binary is too old (from the compilation date), it will display a message box instead of the application.

:m:


fcollot:
The private plugins should handle their own expiration date (or other DRM method), not the core, because one can simply recompile the core and copy the private plugins into it).

Issue #1103, opened 2023-02-13
[medInria4] empty zip/dmg packages on master branch


mathildemerle:
On macOS, compiling medInria 4 on master branch we get:
 * `./superbuild/medInria.sh` works
 * `medInria-4.0.0-x86_64.tar.gz` with only a `bin` directory including `medInria_launcher.sh`
 * `medInria-4.0.0.dmg` with an empty medInria logo
 * (also, there is CMAKE_BUILD_TYPE_medInria which is empty by default, and CMAKE_BUILD_TYPE_ExtProjs is on Debug by default)

If i install the dmg in Applications, and open the content of the application, it's completely empty. 
It seems that it's a copy error. The logs at packaging:

```bash
Run CPack packaging tool...
CPack: Create package using TGZ
CPack: Install projects
CPack: - Install scripts: /Users/mathildemerle/dev/medInria4/medInria-public/build/packaging/apple/ApplePackScript.cmake
CPack: - Install script: /Users/mathildemerle/dev/medInria4/medInria-public/build/packaging/apple/ApplePackScript.cmake
WARNING: Plugin "libqsqlodbc.dylib" uses private API and is not Mac App store compliant.
WARNING: Skip plugin "libqsqlodbc.dylib"
WARNING: Plugin "libqsqlpsql.dylib" uses private API and is not Mac App store compliant.
WARNING: Skip plugin "libqsqlpsql.dylib"
/Users/mathildemerle/dev/medInria4/medInria-public/build/medInria-build/bin/plugins
/Users/mathildemerle/dev/medInria4/medInria-public/build/medInria-build/bin/plugins_legacy
Volume on disk5s2 renamed to medInria 4.0.0
cp: /Volumes/medInria 4.0.0/medInria.app/Contents/MacOS/medInria: Read-only file system
cp: /Volumes/medInria 4.0.0/medInria.app/Contents/PlugIns/libitkDWIBrainMaskCalculatorProcessPlugin.dylib: Read-only file system
cp: /Volumes/medInria 4.0.0/medInria.app/Contents/PlugIns/platforminputcontexts/libqtvirtualkeyboardplugin.dylib: Read-only file system
cp: /Volumes/medInria 4.0.0/medInria.app/Contents/PlugIns/libitkINRDataImageWriterPlugin.dylib: Read-only file system
cp: /Volumes/medInria 4.0.0/medInria.app/Contents/PlugIns/libitkFiltersPlugin.dylib: Read-only file system
[...]

PR #1104, opened 2023-02-14
[macOS] solve package error if packaging failed before

mathildemerle:
Fix https://github.com/medInria/medInria-public/issues/1103
Based on https://github.com/medInria/medInria-public/pull/1096 which is needed

Commit: https://github.com/medInria/medInria-public/commit/bf5665d1acf2a84e8caf9952aa9aca173fd4473f

On macOS, if you stopped a packaging step, or if it failed, the current volume as "/Volumes/medInria 4.0.0" was not removed. That's something i needed to check and remove on MUSICardio Jenkins in a script before the compilation/packaging of the application. But it was not done in medInria.

This commit:
 * removes spaces in volume names since it creates problem in the script under zsh
 * checks previous failed medInria volumes and removes them
 * moves BaseMedinriaPackage.sparseimage.gz in packaging/apple (and not in src/utils/osx_packaging)
 
:m:


Florent2305 (approved):
Ok

PR #1110, opened 2023-03-01
[Launcher] unix - not needed path

mathildemerle:
On Unix (Fedora - Ubuntu - macOS), a path in the launcher was not needed and created log error on Fedora and Ubuntu.

:m:


Florent2305 (approved):
Ok

Issue #1111, opened 2023-03-10
[Packaging] med4 ITKeigen installed on wrong area on Linux master packaging


mathildemerle:
On medInria "master" branch, try on Ubuntu or Fedora to package, and ITKeigen, from ITK is installed in medInria-public_ExtProjs/src instead of the build binary path. 

This problem cannot be found on medInria3.4 on an other version of ITK.

PR #1113, opened 2023-03-21
[ITK] solve itkeigen packaging pb

mathildemerle:
Fix https://github.com/medInria/medInria-public/issues/1111

Itkeigen since ITK5.1.1 had a problem at packaging on Ubuntu at least. It has been discovered during the work of this PR https://github.com/medInria/medInria-public/pull/1112

There were 2 directories at wrong location: include and lib. The include one is solved on line `_eigen3_cmake_install_includedir` and the lib one through `ITK_INSTALL_PACKAGE_DIR`. For the include one we could have used `ITK_INSTALL_INCLUDE_DIR` but unfortunately it does not compile because it affects other part of the code. 

:m:


mathildemerle:
I put this PR as invalid because after some new tests, there is still a problem with ITKeigen on master, that can be particularly seen in RPM package. Indeed, DEB packages warn that there are linked to weird hardcoded ITKeigen locations in the package (/home/etc), but RPM not only warn but block the installation.

ITKeigen is the only ITK ThirdParty module to have this kind of packaging problem, and this could be an ITK bug.

For now, one quick solution is to rebuild the RPM removing the problematic lines as explained [in this doc](https://github.com/medInria/medInria-public/wiki/DEB-RPM-management#if-the-rpm-needs-rebuilding).

The other solution would be to find a proper debug in ITK and/or medInria, as began here. I'm working on it, if i find something i'll tell here.

mathildemerle:
ITKeigen debug done on https://github.com/medInria/medInria-public/pull/1112/commits/ff83d044b0ee16b05fd9d00c18a3c19f5c00638a on PR https://github.com/medInria/medInria-public/pull/1112
I close this PR

PR #1123, opened 2023-04-24
Python packaging and initialization

fcollot:
This PR adds the following:
- adapt to new version of pyncpp
- add packaging code for all platforms
- initialize Python at startup

The packaging needs testing on Linux and Windows.


fcollot:
There was some inconsistency with the name of the pyncpp project (the cmake project was lowercase but the package config files used uppercase i.e. `find_package(PYNCPP)` instead of `find_package(pyncpp)`). This caused a bug in the linux packaging because medInria installs external projects using the same name as the package. I've pushed a PR on pyncpp (https://github.com/LIRYC-IHU/pyncpp/pull/8) and a commit here to fix this.

mathildemerle:
I tried this version of this PR + https://github.com/LIRYC-IHU/pyncpp/pull/9 but on Ubuntu i still have the bug opening ./superbuild/medInria.sh and the package:

`error while loading shared libraries: libpython3.10.so.1.0: cannot open shared object file: No such file or directory`

fcollot:
https://github.com/LIRYC-IHU/pyncpp/pull/9 has been updated and should fix the issue

mathildemerle:
@fcollot compiling, and i can on Ubuntu run medInria from build and package!

mathildemerle (approved):

PR #1124, opened 2023-04-24
[MemoryLeak] medVtkView::buildThumbnail

mathildemerle:
Fix a memory leak building the thumbnail importing a data. This PR removes the need to resize a pop-up to get a thumbnail with the right size. Instead, we resize a QImage.

Tested on Ubuntu and macOS.
 
-> memory leak fixed, no more po-up importing a data, and a quicker import.

:m:

### DIFF
--- a/src/layers/legacy/medCoreLegacy/parameters/medTimeLineParameterL.cpp
+++ b/src/layers/legacy/medCoreLegacy/parameters/medTimeLineParameterL.cpp
@@ -66,6 +66,11 @@ public:
 
         if(numberOfFramesLabel)
             delete numberOfFramesLabel;
+
+        if (extensionShiftParameter)
+        {
+            delete extensionShiftParameter;
+        }
     }
 };
 
@@ -148,6 +153,7 @@ medTimeLineParameterL::medTimeLineParameterL(QString name, QObject *parent):
 medTimeLineParameterL::~medTimeLineParameterL()
 {
     delete d;
+    d = nullptr;
 }
 
 void medTimeLineParameterL::clear()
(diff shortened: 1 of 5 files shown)


src/layers/legacy/medCoreLegacy/parameters/medTimeLineParameterL.cpp
```diff
@@ -153,7 +148,6 @@ medTimeLineParameterL::medTimeLineParameterL(QString name, QObject *parent):
 medTimeLineParameterL::~medTimeLineParameterL()
 {
     delete d;
-    d = nullptr;
```
Florent2305: Keep it

src/layers/legacy/medCoreLegacy/parameters/medTimeLineParameterL.cpp
```diff
@@ -66,11 +66,6 @@ class medTimeLineParameterLPrivate
 
         if(numberOfFramesLabel)
             delete numberOfFramesLabel;
-
```
Florent2305: If you remove these lines and getWidget is never call, ok in normal way this never happens, you create a memory leak.
Because in this case extensionShiftParameter will never parented.
Parent the extensionShiftParameter on **this** during the construction phase is the only good way.

Florent2305 (changes requested):
Sorry but nogo

mathildemerle:
(For info, these comments have been made on an outdated commit, these files changes are not in the final version)

Florent2305 (approved):
I teste this PR on Windows 11 with Qt5.13.2
It's fine for me.
And no objection on code.

PR #1127, opened 2023-05-04
[Release Notes] add pipeline workspace

mathildemerle:
Prepare medInria 3.4 release with the incoming Pipelines workspaces.

To be merged with or after the Pipeline system.

:m:


fcollot:
As said in https://github.com/Inria-Asclepios/music/pull/975 the pipeline workspace is not being ported to the core (forked or not), so the release notes of medInria-public should not mention them.

PR #1131, opened 2023-06-05
[MemoryLeaks] solve several memory leaks -- master

mathildemerle:
Same as https://github.com/medInria/medInria-public/pull/1130 on master.

Fix memory problems in medInria.

:m:


fcollot:
There are some issues here that add unnecessary (and potentially bug-hiding) code.

First of all, related to smart pointers:
- It is not necessary to reset a smart pointer in the destructor as it will be destroyed anyways which will have the exact same effect.
- It is not necessary to initialise smart pointers. They are automatically initialised to contain a null pointer.

Both the issues above are minor, but it adds unneeded lines of code. If pointers are to be initialised and deleted manually, there is no need to use a smart pointer.

The other issue is an old debate, but I would like to state why it is not really a good idea to set a raw member pointer to `nullptr` after it is deleted in the destructor. It may seem like it will prevent some crashes (double deletes, access after destruction etc.), but **this will not work in release builds**. The reason is because when the compiler sees this:
```
medFoo::~medFoo()
{
    delete d->blah;
    d->blah = nullptr;
    delete d;
    d = nullptr;
}
```
it will notice that the new values of `d` and `d->blah` are not used (since they should theoretically not be accessed after the destruction), and therefore it will probably remove these lines for optimisation. This means that, in the (very) unlikely case that these members are accessed after destruction, **the program will crash in release builds but not in debug builds**.
Also in debug builds the compiler will set a deleted pointer to a recognisable value, so that it can detect if a freed pointer is being accessed. Setting it to nullptr removes this safety feature. 
In other words, setting a pointer to `nullptr` in the destructor hides bugs in debug builds and will not make the program more stable in release.

fcollot:
The part about resetting a smart pointer was not quite right in this specific case, as I didn't see it was a static pointer of a singleton, but it is still unnecessary. The singleton will be destroyed by reseting the smart pointer, so if the destructor is being called, then this reset has already taken place.

Florent2305 (approved):
This PR will be simplified them-self after rebase and merge of other PR about MEMORY LEAK.
This PR will conflict with FLEX PR about main.cpp and medMainWindow.cpp but this problem is well known.

PR #1137, opened 2023-06-30
[medDataSources] compilation error on Ubuntu + accent err

mathildemerle:
Solve a compilation error on Ubuntu 20.04 (in getModelIndex lvalue bug), and fix git diff problems with french accents in comments.

:m:

### DIFF
--- a/src/layers/medCore/source/virtualRepresentation/medVirtualRepresentation.cpp
+++ b/src/layers/medCore/source/virtualRepresentation/medVirtualRepresentation.cpp
@@ -790,8 +790,7 @@ void medVirtualRepresentation::removeTooOldEntry()
 QModelIndex medVirtualRepresentation::getModelIndex(medDataIndex index)
 {
     QModelIndex indexRes;
-    QString indexString = index.asString();
-    QString indexAsString = indexToFileSysPath(indexString);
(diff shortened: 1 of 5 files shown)


src/layers/medCore/source/virtualRepresentation/medVirtualRepresentation.cpp
```diff
@@ -790,7 +790,8 @@ void medVirtualRepresentation::removeTooOldEntry()
 QModelIndex medVirtualRepresentation::getModelIndex(medDataIndex index)
 {
     QModelIndex indexRes;
-    QString indexAsString = indexToFileSysPath(index.asString());
```
Florent2305, line 793: Thanks for correction but a better way consist to modify fonctions signatures by
```
QString indexToFileSysPath(const QString &index); 
```
into medDataHub.h

Florent2305 (changes requested):
Thanks for corrections but a little change is required

src/layers/medCore/source/virtualRepresentation/medVirtualRepresentation.cpp
mathildemerle: thx, PR updated with the changes

Florent2305 (approved):
Ok Thx

PR #1141, opened 2023-07-21
make Packaging a dedicated project

fcollot:
There is a major packaging issue on unix that results in the whole superproject being built multiple times (with no optimizations, so in single thread mode) for no reason other than CMake not being able to keep dependency information on install targets with makefile generators (see https://stackoverflow.com/questions/36689963/cpack-tries-to-build-all-targets).
This means the packaging (on MacOS at least) **takes hours instead of minutes for the packaging phase only**.

Example on linux:

Screen Shot 2023-07-21 at 15 41 39

All those _"run preinstall target"_ lines are build commands even though everything has already been built, and from the build times I'm getting it seems to consider everything out of date so is fully rebuilding (worse, the superproject is rebuilt at the very end, so will go through the whole process for the third time...)
(on macOS it only rebuilds once, but still results in a very long packaging phase)

To fix this I had to make the packaging part a separate project so CPack only tries to build the packaging project instead of the superproject. This works on MacOS (takes a few minutes now) but I have to test on Linux to make sure it works (and the gain won't be as great as it will still try to rebuild the external projects, but only once instead of twice).
In addition the code to create the unix launchers has been moved from the packaging to the superbuild, which is conceptually more logical.
On Windows this change will have no effect, but should be tested to make sure it's not broken.


mathildemerle (approved):
Nice, we'll have to test this PR on Ubuntu/Fedora before merging

mathildemerle:
I tried on Ubuntu and doing `make -j8` then `cpack ..`, or directly `make -j8 package` gives only a zip file with `medInria.sh` in some empty directories. There is a problem with this PR, it can't be merged for now

mathildemerle (changes requested):
Does not work on Ubuntu

fcollot:
There was a line missing when I pushed this. It's fixed (see commit), and I'm currently testing on the ci instance but as expected the gain is much smaller on linux and it's taking a very long time.

mathildemerle:
This line fix indeed the almost empty package. 
However, there is a problem in the build directory (i did a fresh compilation): at the root of my build dir, i've got a `medInria.sh` that i can run. But i've got also a `medInria.sh` file in the `superbuild` directory, which do not work. If i run it i've got the error: `./medInria.sh: ligne 49: /home/mathildemerle/Dev/medInria3.4/medInria-public/build/bin/medInria: Aucun fichier ou dossier de ce type`

fcollot:
Before this PR, on Linux, the superbuild configuration code would create two scripts to run the application:

`superbuild/medInria.sh` <-- to be used in the build tree
`packaging/unix/medInria_launcher.sh`  <-- to be packaged with the application

With this PR the script to run the application in the build tree is at the root of the build tree, which is more logical.  I renamed the other one to `medInria.sh` instead of `medInria_launcher.sh` because I think that's more logical too. It is created in the superbuild folder because that's the project that creates it.

mathildemerle:
ok for now i've got after a packaging:
`build/medInria.sh` -> works
`build/superbuild/medInria.sh` -> does not work running it from terminal in build directory, and is not the same file as the first medInria.sh
`build/packaging/linux` -> no script

I don't know if it's a good idea to have 2 files with the same name but which do not do the same thing. However, indeed it's better to run `medInria.sh` than `medInria_launcher.sh` in a linux package. Even if we switch to DEB and RPM on medInria 4 (i'm waiting for the review/merge of my PR), in this system users run only "medInria", so the file change does not change anything for users. 

So, i validate this PR here

fcollot:
These scripts do the same thing (run the executable), but one does it slightly differently because an installed medInria is different to a build tree one.

fcollot:
Just to clarify, I did not change what was done before except rename one version of the script (which as you said yourself is more logical) and also these files are created in different places than before.

Florent2305 (approved):
Some changes look fine, but with cmake we can't be sure without extensive testing.

I validate this PR, but if we find issues we must be very reactive to solve them.

PR #1143, opened 2023-09-05
Shanoir source

LFLCH:
- [x] Connect/disconnect to Shanoir
- [x] Load the tree-view from shanoir
- [x]  View niftii files from Shanoir
- [x]  The plugin is domain name agnostic
- [x] The unzipping process doesn't require a dependency
- [x] Retrieving data from shanoir can be asynchronous
- [x] It is possible to upload data to shanoir
- [x] The upload of new data is asynchronous
- [x] The async process is thread safe
- [x] The errors from the requests to shanoir are managed 
- [x] It is possible to abort a request 
- [x] All the code is documented


Florent2305 (approved):
Great job. I merge it immediately!

PR #1145, opened 2023-09-06
[SplashScreen] speeds up launch removing txt on splashscreen

mathildemerle:
This PR speeds up the launch of the software removing the display of each plugin name on it. 

On my machine (ubuntu) i had a 2 sec long splash screen, and with this i've got a instant launch. On some other machines, the splash screen is way longer so it can help too.

:m:

### DIFF
--- a/src/app/medInria/main.cpp
+++ b/src/app/medInria/main.cpp
@@ -126,8 +126,6 @@ int main(int argc,char* argv[])
     bool show_splash = false;
     #endif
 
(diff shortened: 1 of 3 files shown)


fcollot (approved):
This takes the same time to load on my system except now it just stalls with no text so I'm not sure it's an improvement on macOS, but if it speeds loading on linux ok.

juliencastelneau (approved):

PR #1148, opened 2023-10-02
[VTK] avoid med crash with wrong version of vtk files

mathildemerle:
Same as https://github.com/medInria/medInria-public/pull/1135 on master branch.

If there is a reading problem importing data, medInria should not crash.

:m:


Florent2305 (dismissed):
Ok

mathildemerle:
This code changed between master and dev branch. On dev `medDatabaseReader` files seems to have been removed, and `medDataImporter::convertWithOtherReader` is testing the output of `dataReader->read` method. It needs to be tested with the new system on dev in case we need to do something additionally if the import bug. If it's ok i'll close this PR

Florent2305:
This correction is done on removed file.
Then I close this PR.

PR #1151, opened 2023-10-13
[medVtkImageInfo] proper reset of a complex structure

mathildemerle:
At compilation of medInria i had some warning as:

``` 
medInria-public/src/layers/legacy/medVtkInria/vtkImageView/vtkImage2DDisplay.cpp:67:60: warning: ‘void* memset(void*, int, size_t)’ clearing an object of non-trivial type ‘struct medVtkImageInfo’; use assignment or value-initialization instead [-Wclass-memaccess]
   67 |         memset(&m_sVtkImageInfo, 0, sizeof(m_sVtkImageInfo));
```

m_sVtkImageInfo is a complex structure with constructor and destructor, and should not be reset with memset.

:m:


Florent2305 (approved):
I disagree with Valgrind. But at Mathilde's request and to improve the readability of Valgrind traces, I reluctantly accept.

PR #1162, opened 2023-10-30
[vtkImageInfo] an other way to reset the complex structure and please GCC

mathildemerle:
An other way to be GCC compliant with m_sVtkImageInfo, seen with @Florent2305.

(previously changed on https://github.com/medInria/medInria-public/pull/1151)

:m:


Florent2305 (approved):
A great pleasure to merge it :)

PR #1164, opened 2023-11-06
[medCropToolBox] fix error return value

fcollot:
Minor issue found while reviewing a PR. An internal function using error codes returned `0` (i.e. success) instead of a failure code.
(the error case is never encountered but this is still a technical bug)


mathildemerle (approved):
indeed

Florent2305 (approved):
ok

PR #1167, opened 2023-11-23
[DCMTKreader] roll back on a change on static Qlist

mathildemerle:
Remove these deletions which are on static QLists. It created a crash in MUSICardio side that's why i looked at this.

:m:


fcollot (approved):

Florent2305 (approved):
ok

PR #1168, opened 2023-11-23
[SplashScreen] use directly QSplashScreen & update splash medInria logo

mathildemerle:
As done on MUSICardio side: https://github.com/Inria-Asclepios/medInria-public/pull/800

> This PR removes the use of medSplashScreen which is not needed anymore, we can use directly QSplashScreen.
> 
> It's now launched at the creation of the app (in the beginning of medApplication constructor), and automatically closed when the window of the app is displayed.
> 
> This change debugs also a problem at least of Ubuntu side where the splashScreen was not displayed at the right time.

Also, i updated the medInria splashscreen logo to be a dedicated one (not the same as the one on homepage), with a background color, and no need to space for text.

:m:


Florent2305:
At a moment of code review the question was 
``` 
QSplashScreen *splashScreen
```
leek or not ?

Thanks to Mathilde investigation, the lines
``` 
d->splashScreen->setAttribute(Qt::WA_DeleteOnClose, true);
``` 
``` 
d->splashScreen->finish(d->mainWindow);
``` 
prouve we don't have a leek.

Then for me the PR is good to merge ;)

But a question still open. After the line 
``` 
d->splashScreen->finish(d->mainWindow);
``` 
Do we need a ```d->splashScreen = nullptr; ``` the answer to this question is in a future discussion...

Florent2305 (approved):
Ok

mathildemerle:
as a reminder it's possible to switch `QSplashScreen *splashScreen;` to `QPointer<QSplashScreen> splashScreen;` because QPointer set automatically to null the variable if it has been deleted. It's not really useful here but could be helpful in other part of the code

PR #1174, opened 2023-11-28
[Close] 4D+3D data in view crash at app closing

mathildemerle:
This PR fixes a crash:
 * Open a 4D data in a view, then a 3D mask/data
 * Close the application -> crash

Parameter label was not well removed.

:m:


Florent2305 (approved):
ok after quick direct discution with Mathilde

PR #1176, opened 2023-11-29
[Thumbnail] no need to warn if no thumbn since the thumbnail setting added

mathildemerle:
Since a thumbnail setting has been added to allow or not the creation of the thumbnail at data import, there is no need to display a warning about the existence of thumbnail during the saving of a temporary data.

:m:


Florent2305 (approved):
Ok

PR #1178, opened 2023-12-01
[Histogram] switch QObject+QGraphicsItem to QGraphicsObject

mathildemerle:
From PR https://github.com/Inria-Asclepios/medInria-public/pull/805

> To avoid:
>  * a warning on Ubuntu because of the lack of `Q_INTERFACES(QGraphicsItem)`
>  * a compilation error on macOS because of its existence.
>  
> And to remove an old special comment ( 👀 ).
>  
> This PR changes the double inheritance in some histogram classes to a single one directly derived from QObjects.

:m:


Florent2305 (approved):
QGraphicObject smell really better than combination of QObject and QGraphicItem
Good one

PR #1181, opened 2023-12-06
Remove legacy medPacs. A new plugin for PACS will be developed nearly

juliencastelneau:
Everything is in the title !


Florent2305 (approved):
Good

PR #1183, opened 2024-01-19
[PolygonROI] ROI/repulsor crashs and interactor problems (#807)

mathildemerle:
From PR of MUSICardio https://github.com/Inria-Asclepios/medInria-public/pull/807

> There are some problems with Polygon ROI toolbox.
> 
> ### Solved in this PR
>  * If you want to draw a ROI with SHIFT-click, the first click was not drawing anything. This is important because some users think the toolbox is not working.
>  * A warning message has been solved.
>  * Draw a ROI, activate then deactivate repulsor: user looses the possibility to change the view contrast.
>  * Draw a ROI, activate repulsor, change toolbox (run `clear()` method) then click on the view -> crash.
> 
> ### Problems that are still there
>  * Activate the toolbox, draw 1 landmark, activate Repulsor, click on the view -> crash
>  * The last point fixed above does not crash anymore, however we loose the possibility to change the view contrast. The user needs to choose an other mouse interaction setting and go back to the original one to reset the interactor. It's annoying and it would be nice to solve that, but at least it's not crashing.

:m:

### DIFF
--- a/src/plugins/legacy/polygonRoi/toolboxes/polygonRoiToolBox.cpp
+++ b/src/plugins/legacy/polygonRoi/toolboxes/polygonRoiToolBox.cpp
@@ -395,6 +395,10 @@ void polygonRoiToolBox::clickClosePolygon(bool state)
         {
             viewEventHash[activeDataIndex]->onSelectContainer();
         }
+        else if (!viewEventHash.empty())
+        {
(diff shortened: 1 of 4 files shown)


Florent2305 (approved):
ok

PR #1193, opened 2024-01-22
[Paint] solve problems switching between tools (#770)

mathildemerle:
From PR https://github.com/Inria-Asclepios/medInria-public/pull/770

> This PR solves problems using tools in Paint:
> 
>  * Paint/Erase and Magic Wand buttons do not set properly their widgets.
>  * Clear button was hiding buttons even if data were in the view, run twice a part of the cleaning process, then and Add and Erase buttons when shown were not handled well.

:m:


Florent2305 (approved):
Ok

PR #1207, opened 2024-02-26
[Reslice] background color for selected view

mathildemerle:
As done in https://github.com/Inria-Asclepios/medInria-public/pull/823

This PR adds back a (background) color to display the current selected view.
This is needed for the user to choose the main orientation that is going to be used in the output data.

:m:


Florent2305 (approved):
ok

PR #1209, opened 2024-02-26
[Reslice] background color for selected view --dev

mathildemerle:
Same as https://github.com/medInria/medInria-public/pull/1207 on medInria3.4.x

This PR adds back a (background) color to display the current selected view.
This is needed for the user to choose the main orientation that is going to be used in the output data.

:m:


Florent2305 (approved):
Ok

PR #1233, opened 2024-09-09
[Release3.4] add info about Anima plugins in release notes

mathildemerle:
Due to changes in anima plugins management, and waiting for futur changes in medInria 4 changing the way plugins are handled, Anima plugins are temporarily removed from medInria.

This PR adds a mention about it in the release notes.

:m:


Florent2305 (approved):
Merci

PR #1238, opened 2024-09-23
[Compilation] morphomath_operation ITK modules

mathildemerle:
In order to compile dev branch on macOS and Windows (with VS 2022), we need to add the right ITK module in morphomath_operation plugin.


fcollot:
You can also use `${ITK_LIBRARIES}` which avoids repetition with the components listed in the `find_package` call (this variable lists the targets associated with the requested ITK components).

mathildemerle:
I usually don't use ${ITK_LIBRARIES} for one module but indeed it's possible, i'm busy right now but if needed i'll changed it

Florent2305 (approved):
Ok

PR #1240, opened 2024-09-24
Notification widget bug and minor compilation issue

fcollot:
When the notification popup fades out it is not destroyed. This prevents the user from using the mouse on the area where the popup is still present (but invisible).
(I also fixed a minor compilation issue I encountered while testing testing bug)


Florent2305 (approved):
Ok, very good job.

PR #1249, opened 2024-10-14
[Notifications] fix multiple size problems

mathildemerle:
This PR:
 * fixes height problem, now it ends at the application bottom
 * refreshes size at application resize (buttons or manually)
 * fixes width problem with long text (word wrap)
 * fixes width problem with widget title: display name of the loaded file, and not the path

:m:


Florent2305 (approved):
Ok, I have reviewed all the changes, everything seems correct to me.

PR #1253, opened 2024-10-16
[Annotation] fix A/R/P/L orientation annotations in views

mathildemerle:
On dev branch of medInria, the orientation annotation in center of view borders are not well displayed.

Since newer version of VTK, `vtkCornerAnnotation::SetTextActorsPosition` method takes a `const` value, and if we don't change also in medInria, our virtual `vtkOrientationAnnotation::SetTextActorsPosition` method is not called.

:m:


Florent2305 (approved):
Very good job

PR #1258, opened 2024-12-04
[SplashScreen] display the splashscreen on right screen and in release only

mathildemerle:
Put back the old splashscreen behavior: 
* do not show on debug mode
* display at the previous screen used by the application
* also a small debug: save the current screen even if the app is in fullscreen

:m:


Florent2305 (approved):
Ok

PR #1264, opened 2024-12-18
[macOS] solve missing legendre fct from c++

mathildemerle:
On macOS, the legendre mathematical function is not in std. I tested some solutions but the only valid are : reusing boost, or implementing our own legendre function. This PR adds a legendre method, however i did not test it, but it compiles. If the code is not good enough we could enhance that after the "beta" release.

An other point: QT.._DIR is needed on macOS to compile.

:m:


Florent2305 (approved):
Ok

PR #1268, opened 2024-12-18
Add Python

fcollot:
A port of the Python integration from medInria 4


fcollot:
After testing on Windows with @Florent2305 we changed the code to call the cmake install script manually for pyncpp.

mathildemerle (approved):

